PR #28, opened 2016-10-04
Fix spacing issue on "onlineDRM" article

ahoneybun:


ahoneybun:
Alright I might have been able to do this without a new PR but if this works then I know what to do next time :)

Xeboc:
This could be done with 1 commit, btw.  Make sure to sync your master with upstream master first.  Then create a new branch, and do your edits off that new branch.  That will avoid all of those unnecessary local cleanup commits.  Isn't git fun?  :stuck_out_tongue_closed_eyes:

PR #37, opened 2016-11-19
Update switch.md

rdfuhr:
Change the term "Apple's taskbar" to "Apple's dock"


Xeboc:
Ah, nice catch.  Thank you!

PR #52, opened 2017-05-15
add clicking noise fix via pulseaudio config clearing

EvanKepner:
Issue I had with a new Lemur with the resolution.


Xeboc:
Solid.  Thanks for adding to the document!

PR #55, opened 2017-05-31
Fix md rendering

mrjoshida:
Page was not rendering properly due to indentation and markdown syntax issues.


Xeboc:
Nice fix.  Thank you for cleaning that up.  Appreciated!

PR #57, opened 2017-06-30
Minetest docs and other FOSS

kylerschin:


Xeboc:
Kyler!  Thanks!  You rock!

PR #58, opened 2017-07-05
Pull styles out

cassidyjames:
Getting the styles into Artoo. This is a start to possibly switching these templates to take advantage of the v2 styling as well.


_layouts/default.html
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html class="no-js">
   {% include head.html %}
-  <body>
+  <body class="path-support">
```
bencates: Should we pull in a feature-parity `path-` helper? We know the full path at build time, and I know it's possible in Jekyll (though maybe doing it withing GH Pages sandbox rules might be tricky).

cassidyjames: We _could_ but I don't think I actually want to. I wanted to scope the whole site to these styles (since that's where the styles lived in Artoo), but I don't think I actually want the ability/responsibility to have per-article styles.

.gitignore
```diff
@@ -1,3 +1,5 @@
 _site
 .sass-cache
 .live-archive/
+Gemfile.lock
```
bencates: NO! BAD!

This should only change when the dependencies change, which is a code change and should be in source control. Keeping this file in the repo eliminates a huge class of version mismatch bugs.

NPM (pre-5) is teaching bad habits.

PR #61, opened 2017-09-07
rephrased second paragraph's second sentence a bit for clarity.

ahoneybun:
On the note about the link to OEM Firstboot, is that a link to a future article or just left over from something else?


Xeboc:
Thanks Aaron, appreciate the edit.  That document it references is a current document:

http://support.system76.com/articles/oem-firstboot/

PR #76, opened 2018-02-28
Use Denver time

btkostner:


isantop:
"Mountain Time" is both more accurate and likely more helpful than "Denver Time" in case people don't know which time zone Denver is in.

btkostner:
This was something Ben pointed out. Ask him.

isantop:
I think we all agree here (@brs17 should comment so he's on record).

brs17:
Had pointed out MDT should be Mountain time on the contact page. Fixed here: https://github.com/system76/joshua/commit/ececa2e73311f2451fb27fa0449488b7b03ad584

PR #81, opened 2018-04-19
Upgrade Pop

Socialhappiness:


_articles/upgrade-pop.md
```diff
@@ -0,0 +1,78 @@
+---
+layout: article
+title: Upgrade to Pop!_OS 18.04 LTS
```
Xeboc: This should be:

Pop - Upgrade Pop!_OS

It doesn't need to be to a specific version, and we have a single word that leads the title.

_articles/upgrade-pop.md
```diff
@@ -0,0 +1,78 @@
...
+  - Update
+  - Release
+  - System76
+  - apt
+image: http://support.system76.com/images/upgrade-ubuntu/Ubuntu-17-04.jpg
+hidden: false
+section: articles
+
+---
+To install Pop!_OS for the first time or get the new recovery system (recommended), download and install from https://system76.com/pop. Though not necessary, we recommend a fresh install with this release for the recovery system. In future releases the recovery partition will download the next version enabling more reliable on-disk release upgrades. The recovery partition also provides an always-stable live environment that helps speed up diagnosing and fixing support issues.
```
Xeboc: This could be rewritten for clarity:

Pop 18.04 includes a new recovery system feature.  If you would like this new feature, please reinstall the operating system as described here: http://support.system76.com/articles/install-pop/.  Future upgrades will update the recovery partition enabling more reliable upgrades. The recovery partition also provides a live environment that helps speed up diagnosing and fixing support issues.  Upgrading in place will not create a recovery partition.

_articles/upgrade-pop.md
```diff
@@ -0,0 +1,78 @@
...
+image: http://support.system76.com/images/upgrade-ubuntu/Ubuntu-17-04.jpg
+hidden: false
+section: articles
+
+---
+To install Pop!_OS for the first time or get the new recovery system (recommended), download and install from https://system76.com/pop. Though not necessary, we recommend a fresh install with this release for the recovery system. In future releases the recovery partition will download the next version enabling more reliable on-disk release upgrades. The recovery partition also provides an always-stable live environment that helps speed up diagnosing and fixing support issues. 
+
+### Create Installation USB
+
+First, we recommend having Pop!_OS installation media handy before starting any upgrades in case something goes wrong. You can find instructions on creating a a USB with Pop!_OS from [support.system76.com](http://support.system76.com/articles/live-disk/).
```
Xeboc: Use the relative link to allow for non-breaking site changes:

/articles/live-disk/

_articles/upgrade-pop.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Backup Your Files
+
+Next, we recommend backing up all important files from your computer. The upgrade process will leave your files intact, but it's best to be safe in case something goes wrong. Please read our article on [How to Backup Your Files](articles/backup-files/) for helpful instructions.
```
Xeboc: this link needs to be:

/articles/backup-files/

_articles/upgrade-pop.md
```diff
@@ -0,0 +1,78 @@
...
+
+Once restarted the computer will be on the newly upgraded system! If you run into any issues, check out Troubleshooting below.
+
+## Troubleshooting
+
+Most upgrades proceed without a hitch, but occasionally things go wrong. If your system hangs on the upgrade process or if it reboots and won't start, then the following steps can help repair your broken OS.
+
+### Stuck Upgrades
+
+If your upgrade appears to hang in place for an extended period of time, click on the **Terminal** item to expand the terminal section. Check what action is available there, then complete the steps to unhang your upgrade. For help, contact support and we'll be able to provide assistance.
```
Xeboc: We should mention that there will probably be questions about 'use package maintainers version', and that it's pretty safe to choose 'y'.  Choosing 'y' won't break anything, so we want to encourage people to choose 'y'.

_articles/upgrade-pop.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Broken Upgrade
+
+Please see the steps above to create a bootable USB. Restart your computer and start your computer from the bootable USB. When given the option select your Language and Keyboard then choose **Try Demo Mode**.
```
Xeboc: We don't want to encourage people to reboot if the `do-release-upgrade` fails.  The operating system is still running, and it's much better to fix it while it's still running.  A reboot is a bad idea.  Most likely, they just need to run `do-release-upgrade` again.  At worst, they need to do the 6 package manager commands and start repairing the package system manually.  We really should repair it while its still working, it's much easier.

_articles/upgrade-pop.md
```diff
@@ -0,0 +1,78 @@
...
+
+Next follow the steps to repair grub first:
+
+[Repair Grub](/articles/grub/)
```
Xeboc: This won't apply if they have a fresh install of Pop 18.04, since they won't have GRUB.  This won't be accurate on the next upgrade.  If we have this, we should say 17.10 - do GRUB, 18.04 - do systemd-boot repair. (document not yet created).

_articles/upgrade-pop.md
```diff
@@ -0,0 +1,78 @@
...
+
+Type the following command, followed by the enter key.
+
+```
+sudo do-release-upgrade
+```
+
+You'll be prompted to enter your system password, but when you type it, the letters won't show. Just continue typing the password and press <kbd>Enter</kbd>
+
+Type <kbd>Y</kbd> and press enter when prompted
```
Xeboc: There will be 3 prompts.  One to install packages, one to remove packages, and one to reboot.  There is also a keyboard layout prompt in some situations.  We should mention that they need to watch the upgrade window and choose 'y' for the 3 prompts.

_articles/upgrade-pop.md
```diff
@@ -0,0 +1,78 @@
...
+keywords:
+  - Pop!_OS 17.10
+  - Pop!_OS 18.04
+  - LTS
+  - Upgrade
+  - Update
+  - Release
+  - System76
+  - apt
+image: http://support.system76.com/images/upgrade-ubuntu/Ubuntu-17-04.jpg
```
Xeboc: We will want a different image for Pop.

_articles/upgrade-pop.md
ahoneybun: I've seen things break when not picking local version. I believe this applies to config files of applications like apache and mysql.

_articles/upgrade-pop.md
ahoneybun: 3 'y' prompts and one enter.

_articles/upgrade-pop.md
Xeboc: We could phrase it as:

"if you haven't specifically made a change to a configuration file in your system, go ahead and press 'y' to use the package maintaner's version.  If you have made a change you would like to keep, press 'n' to use the local version, or press 'd' to inspect the changes and see which version you would like to use."

PR #91, opened 2018-06-11
include pepperflashplugin-nonfree on install

simonsjw:
I run chromium on pop os and couldn't view a flash video. I tried installing browser-plugin-freshplayer-pepperflash on it's own and it didn't do anything. Installing the additional non free plugin seemed to do the trick. The advise at 
https://askubuntu.com/questions/836261/pepperflashplugin-nonfree-package-installation-fails-since-chrome-54-is-out-oct
is to not use this package at all and instead use adobe-flashplugin. I've got it working for now so will trade the resultant vulnerabilities for having it work. However, once I'm out of exam season, I'll revisit. 
Simon

### DIFF
--- a/_articles/flash.md
+++ b/_articles/flash.md
@@ -28,7 +28,8 @@ section: articles
 To watch Flash videos with <u>Firefox</u>, <u>Chromium</u>, and <u>Opera</u>, we recommend using <u>Pepperflash</u>, rather than <u>Adobe Flash</u>, for security reasons.  <u>Pepperflash</u> is maintained by Google and included in the <u>Chrome</u> browser by default, and can be downloaded separately to be used in other browsers.  Please run this command to install <u>Pepperflash</u>:
 
 ```
-sudo apt install browser-plugin-freshplayer-pepperflash
+sudo apt install browser-plugin-freshplayer-pepperflash pepperflashplugin-nonfree
+
 ```
 
 Otherwise, to install <u>Google Chrome</u>, follow these instructions:


Xeboc:
Thank you for the addition.  Can you provide the website?  I'd like to test on our end.  I'm not sure we want to recommend using something that is considered more vulnerable, but I'd like to look into the best ways to use Flash on Pop.

simonsjw:
Hey James,
Thanks - it's great that you are testing. If I could live without Flash I so would! Congrats on Pop by the way - I'm a data scientist and I've come over from Arch to focus on the data rather than the setup. I use some Containers via nvidia-docker on a sytem76 Onkyo and the whole thing just hums along.

The warning was on the link I sent previously to askubuntu. The advice on install is from the below. It appears to be about running Flash on a system76 system but using Manjaro. 

https://manjaro.site/install-pepper-flash-plugin-pop-os/

Do let me know how you go on. 

Simon

ahoneybun:
I'm sorry for the delay on reviewing this PR. Did you restart Chromium after installing the browser plugin or did you restart the browser after installing both packages?

simonsjw:
Hi Aaron,
I did. In the end I switched to Chrome though - that already has the flash
plugin installed. I'm happy to help with testing though. I've clean
re-installed a number of times since this ticket but if you send details of
what you would like me to try (if anything), I'll happily give it a go.

Thanks for getting back to me.

Simon

ahoneybun:
I've gotten websites like Pandora to work in Firefox by installing the single package that we have listed currently. Is there other websites that need the other package to work correctly?

ahoneybun:
@simonsjw Did you hit any websites that didn't work without this package? Thank you.

simonsjw:
Hi Aaron - funny you should get in touch. 
For some reason I can't use 
[https://www.poptropica.com/game/](url)
on Chrome. (Flash is definitely activated)
I previously could run this site after activating flash. 
Simon
Screenshot from 2019-05-18 13-12-34

ahoneybun:
I think that the package was removed from 19.04 at least as I can't seem to find it in the archive.

mmstick:
You could try getting the deb from Adobe's website: https://get.adobe.com/flashplayer/

Though it won't be much longer until Adobe finally pulls the plug entirely and stops hosting flash player entirely.

PR #109, opened 2018-08-23
Fix typo in Ubuntu cuda install instructions

trxcllnt:
- Append apt source entry to `sources.list`, not `source.list`
- Remove `/` between bionic and main, because apt complains
- Specify `arch=amd64`; not sure if that'll work for everyone, but was necessary for me on a 2018 Bonobo

Unrelated question: should the `apt.pop-os.org` domain be served over HTTPS?


trxcllnt:
@jackpot51 :pray:

PR #118, opened 2018-09-16
Sharing command lines were inversed

Jiab77:
Just moved the command lines in the correct places


jackpot51:
Thanks!

Jiab77:
> Thanks!

You're welcome glad I could help

PR #119, opened 2018-10-07
Minor grammar fixes

trinitrot0luene:
Just some very minor omissions that I noticed while reading through the article.


btkostner:
Thanks for the PR!

PR #123, opened 2018-10-31
fix(upgrade): add note about `pop-desktop` missing dep

alexsasharegan:
Hello! I considered doing this as an issue, but this seemed easy enough. I'm going through the release upgrade process, and I ran into an unmet dep for `pop-desktop`. It's a simple fix to install the missing package, but this might be something you want your package to include when it is installed (I don't really know how apt packages trigger sub-dep installs).

So far this is the only deviation I've encountered from the upgrade docs. Thank you!

### DIFF
--- a/_articles/upgrade-pop.md
+++ b/_articles/upgrade-pop.md
@@ -49,6 +49,12 @@ Before beginning the upgrade, it is important to ensure that the `pop-desktop` m
 sudo apt install pop-desktop
 ```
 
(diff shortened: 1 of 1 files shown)


btkostner:
@alexsasharegan Thank you for the PR!

alexsasharegan:
You're welcome! Thank you for solid upgrade instructions! I'm up and running on 18.10 without any issues! I had the encryption setup, so fingers always crossed haha

PR #131, opened 2019-01-13
New dual booting

ahoneybun:


_articles/pop-live-disk.md
```diff
@@ -29,7 +29,7 @@ A live disk is a handy tool to have around!
 
 ## Make Bootable drive
 
-In order to install Pop!_OS, we must first download the .iso image, this is a disk image with the operating system and installer on it. You can download [Pop!_OS here](http://pop.system76.com), just click the link and the download should begin!
+In order to install Pop!_OS, we must first download the .iso image, this is a disk image with the operating system and installer on it. You can download Pop!_OS [here](http://pop.system76.com), just click the link and the download should begin!
```
thomas-zimmerman: The "here" target is a bit too small, I'd suggest it be with "download Pop!_OS here".

ahoneybun: That comment is about the pop-live-disk article not this one.

PR #134, opened 2019-02-12
Fix Sources file in TENSORFLOW docs

cassidyjames:
Correct a typo in the line plus add it to its own file to prevent messing up the main list.


ahoneybun:
Thanks!

PR #136, opened 2019-02-18
Added section about Ubuntu boot-repair

truefakts:
My oryx pro is physically out of arms reach while in operation at my desk. I can however see the screen. The boot-repair utility allows me to select my OS via the grub menu

POP_OS! Also sometimes doesn't detect the presence of Windows 10. This utility is extremely user friendly, and a sure-fire way to fix that problem

### DIFF
--- a/_articles/dual-booting.md
+++ b/_articles/dual-booting.md
@@ -58,6 +58,30 @@ Once you have your partition(s) selected, select the red "Erase and Install" but
 
 Choose your previous OS with the arrow keys, then press Enter.
 
+## Operating System Selection on Startup
+
+In some instances, POP_OS! may not automatically see the presence of other operating systems on your machine. Or simply it may be undesirable to hold the spacebar on boot to be presented with a boot menu. An alternative approach to using the built in EFI boot selection menu, is to install grub using the Ubuntu Boot Repair utility which can be found here:
+
+```
+https://help.ubuntu.com/community/Boot-Repair
+```
+To use this utility, follow these steps:
+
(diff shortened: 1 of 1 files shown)


ahoneybun:
The issue here is that Pop doesn't use GRUB for booting and I think boot-repair is not being worker on.

truefakts:
> The issue here is that Pop doesn't use GRUB for booting and I think boot-repair is not being worker on.

Fair enough! https://sourceforge.net/p/boot-repair/home/Home/ the tutorial on the Ubuntu page cites Unetbootin, which is definitely dead. And source code doesn't look like it has been updated in 5 years.

It does work with 18.04/18.10 on Oryx Pros at least! But totally understandable if you don't want to put that on your official support page. Would a community page be more appropriate then? Since there is a large warning stating that those articles are not curated by System76, but instead are handy tips the community who uses your products have found.

Otherwise it can be rejected all together, I'm cool with that. I just found it handy on my end when dual booting

ahoneybun:
It may be useful for a community article but the boot-repair software is still not being updated. It's also that we're working on an updated version of this article which hasn't been pushed yet.

truefakts:
This pull request can be rejected then. Looking forward to seeing what you guys produce! The current page recommends holding the space-bar to boot into Windows, which is hit or miss at this time

ahoneybun:
It shouldn't be a hit or miss if your using systemd-boot. Tapping ESC also works for systemd-boot and GRUB.

Aaron Honeycutt
System76, Inc.
Customer Support Tech
aaron@system76.com

PR #137, opened 2019-02-18
Updated FlexiCharger article

truefakts:
The original article wasn't very clear as to what flexicharger actually does. This setting is extremely handy, and unique to system76/Clevo in my experience. Power users looking for desktop replacements surely wish for something like this, but may be unfamiliar with how it works.


ahoneybun:
LGTM. Thanks!

PR #139, opened 2019-02-20
Publishing CodeWeavers  case study with James Ramey

kenhara76:


kenhara76:
My thought was that the arrow indicates it is going to be downloaded onto their machine (which is what happens when clicking the link), whereas the right arrow indicates it's going to change their window when they click on it. 

A few examples where this is the case elsewhere:

screenshot from 2019-02-20 10-37-20

screenshot from 2019-02-20 10-35-39

screenshot from 2019-02-20 10-36-59

That said, I can submit a separate PR changing this across all of the case studies if you prefer?

kenhara76:
In fact, we could add an underline below the arrow to make it even more consistent with these examples?

ahoneybun:
Fair enough.

ahoneybun:
So it looks like the PDF for that case study is not in the folder so I get a 404 when I click on that link.

PR #141, opened 2019-02-25
Customize GNOME - Propose a Correction

PTrottier:
Replace "updates" with "updated", it sounds better.


ahoneybun:
LGTM! Thanks!

PR #143, opened 2019-02-27
apt install default-jdk

swaroopch:
`default-jdk` will take care of the right version to install.


ahoneybun:
It looks like that's a metapackage and on 18.10 it pulls in version 11. Not sure if that will break anything in that article though. Have you tested that with this article?

swaroopch:
@ahoneybun Yes, the demo run in Minecraft works for me.

swaroopch:
@ahoneybun I'll play around with this more and resubmit if it looks fine.

PR #144, opened 2019-02-28
Fix typo in laptop-firmware.md

jordanbancino:
“you’re laptop firmware” -> “your laptop firmware”

This is just a small simple thing, but it stuck out when I read it.


btkostner:
@jordanbancino 	Thank you for the PR!

PR #147, opened 2019-03-19
Add article Login with Yubikey

71nuxadd1ct:


71nuxadd1ct:
> I have a few notes when we tested this here:
> Need to install `yubikey-personalization` for the `ykpersonalize` command
> Useful to know how to have a backup root session (`sudo -i`) open.

I updated the apt dependencies, but I already added a line about opening a sudo terminal before modifying PAM.

If you think it's warranted, I could make it more noticeable/add commands on how to accomplish it.

ahoneybun:
LGTM! It looks really good! Thank you!

PR #148, opened 2019-03-24
Update pop-live-disk.md

sgmarc:
"Fixing the boot loader" path incorrect. Fix, please.


ahoneybun:
Thank you for the good work!

PR #150, opened 2019-04-01
Add article: Two-factor FDE with Yubikey.

71nuxadd1ct:


ahoneybun:
It looks like the last commit deleted the file that you were trying to merge, is that correct?

71nuxadd1ct:
I had originally named the file yubikey-fde.md, I created another one called yubikey_fde.md to match the previous article I submitted a pull request for: yubikey_login.md.
So, yes.  The last commit did delete a file, but it was a duplicate.

ahoneybun:
Ah makes perfect sense.

ahoneybun:
I haven't been able to get this to work for me. It's not taking the password that I set with the last command with the Yubikey plugged in. Is there a certain version of Yubikey that this needs?

71nuxadd1ct:
I've updated the guide with your suggestions.
I don't know why it's not working for you.  Are you trying it in a VM?  It works on my Galago Pro (bare metal).

71nuxadd1ct:
If you're testing this in a VM, make sure the Yubikey is being passed as a USB device prior to modifying the LUKS header.

I just finished testing in VirtualBox 6.0.2 r128162, and I had to install the guest-additions-utils and guest-additions-dkms packages (I se Oracle's VirtualBox repo, not the packages in the default repo) and then add the Yubico USB device to the VM.

ahoneybun:
I wasn't using a VM and instead using my personal Galago Pro with Pop 19.04 and wasn't able to login with the Yubikey. I tried two different slots with different passwords as well.

71nuxadd1ct:
Is the 19.04 ISO available for download anywhere?  If so, I can do some testing.  Otherwise, I know it works with 18.04, 18.10, and the 17.04 release.
I can modify the opening of the guide to reflect that it works only for those versions if need be.

ahoneybun:
I upgraded to 19.04 manually. We do have an ISO internally for testing. 

Aaron Honeycutt
System76, Inc.
Customer Support Technician
aaron@system76.com

71nuxadd1ct:
OK.  I've updated the opening of the guide to reflect that it is known to work with versions 17.10, 18.04, and 18.10.
If you can share the ISO, I can test it with the later version.

71nuxadd1ct:
I think it's working on 19.04 now.  Looks like the cryptsetup in the 19.04 release uses a different LUKS header (version 2.0?), and whatever it's trying to use by default to unlock the partition doesn't work.  But, the package comes with a 'ykluks-keyscript' in '/sbin' that works.

Updating the kernel boot options to use this script for unlocking appears to work.

doomspork:
Yikes! Sorry for the trouble @71nuxadd1ct, we still want your contributions! Could you please retarget your PR to the `master` branch instead of `gh-pages`? That should be all that's needed to get this reopened.

PR #151, opened 2019-04-04
Fix it's typos to its

craigstevenlaw:
Hi - wanted to try contributing and so this is my first attempt at that. I just went though the docs to see where we had the incorrect it's and fixed them to the possessive its where appropriate.


ahoneybun:
LGTM to me! Thank you for the contribution and here's to many more!

PR #161, opened 2019-05-30
Update cuda.md

ericphanson:
I didn't see a `system76-cudnn-latest`, but this should now be version `10.1` instead of `10.0` otherwise if you run the first command (`sudo apt install system76-cuda-latest`, which currently installs cuda 10.1) then when you run `sudo apt install system76-cudnn-10.0` it additionally installs the whole cuda 10.0 (after you just installed 10.1).

### DIFF
--- a/_articles/cuda.md
+++ b/_articles/cuda.md
@@ -25,7 +25,7 @@ sudo apt install system76-cuda-latest
 To install the cuDNN library, please run this command:
 
 ```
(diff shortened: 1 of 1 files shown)


kylerchin:
I approve your changes lol

ericphanson:
kind of annoying having this open forever, so I'm just gonna close. Presumably you don't need this PR to make the 1 character change.

PR #174, opened 2019-07-16
Changed KDE Installation

Tiixxel:
Changed the kubuntu-desktop Line to kde-standard because kubuntu-desktop installs Kubuntu's KDE and not the normal KDE Plasma.


thomas-zimmerman:
We recommend the kubuntu desktop as there is a team behind that meta package. The are also the expected set of KDE programs and tools installed with the `kubuntu-desktop` meta package. I don't think there are strong feeling either way.

PR #180, opened 2019-08-12
Update dual boot

ahoneybun:
Should fix this issue once merged to the Pop page as well:
https://github.com/pop-os/docs/issues/81


leventerevesz:
Thank you for writing these paragraphs, they convey the information clearly.

There are two misspelled occurances of "its" in the added text, could you fix them please?

ahoneybun:
Can you point those out?

PR #183, opened 2019-08-16
add Pop 18.10 section

ahoneybun:


ahoneybun:
Should be good now @jackpot51 .

PR #186, opened 2019-08-23
Update desktop-quickstart.md

angerber:
Making it more apparent that if the encryption password is lost, the drive password is lost.


ahoneybun:
Writing down a password is a really bad idea from a security POV.

brs17:
> Writing down a password is a really bad idea from a security POV.

That is a fair point, but the rest of the sentence/changes are still valid, don't you think? It probably is a good idea to mention that there is nothing that can be done if the password is lost/forgotten.

jacobgkau:
The article already contained (and still contains) this sentence, despite Aaron calling it a really bad idea:

"It’s important to write this password down and keep it somewhere safe."

PR #197, opened 2019-09-18
Add article for joining Pop!_OS as an Active Directory client

jacobgkau:
This article explains how to set Pop!_OS up to authenticate with Active Directory credentials (which someone opened a support ticket to get working.)


mmiat:
I use [PBIS](https://github.com/BeyondTrust/pbis-open/releases) to join to AD
does PopOS have an integrated app instead?

ahoneybun:
It looks like Ubuntu doesn't package PBIS so there isn't an integrated application for this.

PR #202, opened 2019-09-28
Install KDE in PopOS!

jwackito:
Instructions to install KDE desktop with the same functionality as in PopOS! with Gnome.


ahoneybun:
Thank you for the article though we already have an article that goes over different desktops and one that goes over switching graphics that includes the command line options as well:

https://support.system76.com/articles/desktop-environment/

https://support.system76.com/articles/graphics-switch-ubuntu/

PR #208, opened 2019-10-15
Tensorman

ahoneybun:
Don't merge until Pop 19.10 is released.


jackpot51:
Is this obsoleted by https://github.com/system76/docs/pull/211?

ahoneybun:
I believe we would merge this first and then #211 .

jackpot51:
#211 contains the same changes as this, so why would we need to merge this?

ahoneybun:
@jackpot51 I *believed* it was. If this is not the case then we can just merge the other one.

jackpot51:
Let's see what happens when #211 is merged. If anything appears missing we can revisit this one

PR #211, opened 2019-10-19
Tensorman File Edited for Content

nathanielsystem76:
Proofread and reviewed by Aaron and I.

### DIFF
--- a/_articles/tensorman.md
+++ b/_articles/tensorman.md
@@ -1,0 +1,151 @@
+---
+layout: article
+title: Use Tensorman
+description: >
+  Using TensorMan to manage TensorFlow Containers
+keywords:
+  - NVIDIA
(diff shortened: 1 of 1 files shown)


_articles/tensorman.md
```diff
@@ -0,0 +1,151 @@
...
+sudo apt install nvidia-container-runtime
+```
+
+The user account working with Tensorman must be added to the `docker` group if that hasn't been done already:
+
+```
+sudo usermod -aG docker $USER
+```
+
+If Docker was just installed, log out and back in before using Tensorman.
```
jackpot51: A log out may not be enough on 19.10, due to systemd user sessions. Reboot is better.

ahoneybun: Fixed.

PR #218, opened 2019-10-22
Fix spelling & grammar inconsistencies on the graphics-switch pages (Pop_OS! & Ubuntu)

trinitrotoluene:


ahoneybun:
You'll need to rebase this on the newest pages since there was an update to those pages for the new Hybrid mode.

jacobgkau:
I really appreciate this PR, as I think we should put more of a focus on grammar/style than we do sometimes. However, these articles have been rewritten, and this PR no longer works.

I have submitted PR #246 against the current versions of the articles, and I'm closing this PR since it is obsolete.

PR #221, opened 2019-10-24
Create KDE-Sys76-Graphics.md

JerryK-1291:
This is for helpling people using a plasma desktop to have a gui process to change graphics cards.

### DIFF
--- a/_articles/KDE-Sys76-Graphics.md
+++ b/_articles/KDE-Sys76-Graphics.md
@@ -1,0 +1,100 @@
+---
+layout: article
+title: GUI Scripts for KDE Desktops
+description: >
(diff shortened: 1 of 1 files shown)


JerryK-1291:
If this was just for a plasma desktop on POP OS I would agree.  However as these also apply to people that pt Kubuntu or KDE Neon, etc on a msystem76 machine would it not be better to have in same location as instructions on how to install the driver?

Thanks

ahoneybun:
The same steps that someone using Kubuntu or Ubuntu/Pop + Plasma would apply to KDE Neon as well so I don't think this article fits on the support pages.

PR #229, opened 2019-11-06
tensorman: Add section for new image creation feature

mmstick:


_articles/use-tensorman.md
```diff
@@ -150,6 +150,44 @@ To find installed containers the `list` command can be used:
...
+
+```
+tensorman =IMAGE_NAME run --gpu bash
+```
+
+> The `--python3` and `--jupyter` flags do nothing for custom containers, but `--gpu` is required to enable runtime support for the GPU.
+
+## Removing the custom image
+
+Images saved through tensorman are manageable through tensorman. Listing and removing works the same:
```
ahoneybun: Needs the underline for tensorman.

PR #231, opened 2019-11-09
Update docking station page to reflect my experience with the HP G2 Dock on the Galago Pro

amdi:
Just received my Galago Pro on Wednesday. I successfully connected it both at home and at work to my HP G2 Thunderbolt Docks with 2 1080p monitors, 3 usb ports, and ethernet. All I had to do was authorize the docks in the Pop OS Thunderbolt menu. I assume it relies on the dkms driver but that was already installed. Worked much more seamlessly than I anticipated. I'm extremely pleased.

Issues:

- ~~Occasionally gets finnicky and won't recognize the attached monitors especially when resuming from suspend with the Thunderbolt port plugged in. I tend to just unplug the thunderbolt port when suspending and plug it back in after I've resumed from suspend.~~
- Seemed to have trouble driving more than 3 USB ports. When I added a daisy chained a USB hub & tried to drive my cooling pad's fan in addition to the above devices it would no longer drive the monitors.
- ~~Can't wake the laptop from suspend with the lid closed. I have to open the lid, wake the laptop, and then the monitors are recognized properly. However, once the monitors are connected it works fine. Any ideas for fixing that would be appreciated (I assume there's some linux system setting that would allow me to unsuspend with the lid closed. I'll probably investigate this eventually)~~

Still, driving multiple monitors, using USB devices, and charging from a single port is more than I expected since I've had bad luck with these Thunderbolt docks in the past. 

Edit: The two issues I crossed out were fixed simply by enabling the wakeup flag on my USB devices, e.g. for all USB devices `echo enabled|sudo tee /sys/bus/usb/devices/*/power/wakeup`

### DIFF
--- a/_articles/use-docking-station.md
+++ b/_articles/use-docking-station.md
@@ -27,8 +27,8 @@ We have tested the following docks:
 ### Community-tested docks:
 
 Community members have reported that the following docks work with our products:
- - [Dell WD19TB Thunderbolt Dock](https://www.dell.com/en-us/work/shop/dell-thunderbolt-dock-wd19tb/apd/210-arik/pc-accessories) [community-tested on an Intel system](https://github.com/system76/docs/pull/206) (**)
- - [HP Thunderbolt Dock 120W G2](https://www.amazon.com/gp/product/B07DPKVYXR/ref=ppx_yo_dt_b_asin_title_o00_s01?ie=UTF8&psc=1) [community-tested on an Intel system](https://github.com/system76/docs/pull/231)
+ - [Dell WD19TB Thunderbolt Dock](https://www.dell.com/en-us/work/shop/dell-thunderbolt-dock-wd19tb/apd/210-arik/pc-accessories) [[community-tested](https://github.com/system76/docs/pull/206) on an Intel system] (**)
+ - [HP Thunderbolt Dock 120W G2](https://www.amazon.com/gp/product/B07DPKVYXR/ref=ppx_yo_dt_b_asin_title_o00_s01?ie=UTF8&psc=1) [[community-tested](https://github.com/system76/docs/pull/231) on an Intel system]
 
 ### For Intel systems
 
--- a/_articles/use-docking-station.md
+++ b/_articles/use-docking-station.md
@@ -28,7 +28,7 @@ We have tested the following docks:
 
 Community members have reported that the following docks work with our products:
  - [Dell WD19TB Thunderbolt Dock](https://www.dell.com/en-us/work/shop/dell-thunderbolt-dock-wd19tb/apd/210-arik/pc-accessories) [Community-tested on an Intel system] (**)
- - [HP Thunderbolt Dock 120W G2](https://www.amazon.com/gp/product/B07DPKVYXR/ref=ppx_yo_dt_b_asin_title_o00_s01?ie=UTF8&psc=1) [Community-tested on a galp4 Galago Pro (2019 model). Drives two 27" 1920x1080 DisplayPort monitors, charging, ethernet, and 3 USB connections simultaneously through the single Thunderbolt port on the Galago Pro.] (**)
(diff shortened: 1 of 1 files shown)


qdrop17:
Hey @amdi & @ahoneybun.

Sorry I might hijack this issue a bit here. I'm having quite some issues with  10th-Gen notebook (in my case running Ubuntu 19.10) when it comes to this docking station:

https://askubuntu.com/questions/1206441/usb-devices-and-nic-not-detected-when-hotplugging-thunderbolt-3-intel-icelake

Is PopOS using different drivers / kernels than Ubuntu 19.10 in regards to TB3 compatibility? I'm dying to use Linux on this device - yet I didn't manage to get this dockingstation running properly yet. Maybe you have some valuable input for me. Any help is highly appreciated.

However: I'm going to try PopOS tonight / tomorrow.

jacobgkau:
@qdrop17 This support article's purpose is to list compatible docks for System76 products. It sounds like you're using the dock with a non-System76 product?

The kernel that ships in Pop!\_OS can be found here: https://github.com/pop-os/linux 

The kernel is 5.3.0 (the same version Ubuntu 19.10 ships by default) at this time, and is based on the Ubuntu 19.10 kernel. You can see the changes that have been made in the repo I linked above. I don't believe Pop's kernel will do anything differently than Ubuntu's for Thunderbolt compatibility right now, but you can look through the repo to be sure.

bradleesand:
I just bought this dock to try to connect two monitors to my Galago Pro and it's not working. I authorized the thunderbolt connection and installed the display port drivers and my mouse and keyboard work when I connect them to the USB ports of the dock, but my monitors aren't being recognized. Any suggestions? I've both rebooted and done a full shutdown/power on cycle and still nothing.

amdi:
I didn't have to install any drivers on latest PopOS. I have 2 1080 monitors hooked up with a displayport -> displayport cable. Sorry I can't think of any suggestions other than that. It all works out of the box for me.

helpme789:
Hi,
I am not sure if this is all dock related, however, I followed your recommendation to change my USB's power on wake settings "echo enabled|sudo tee /sys/bus/usb/devices/*/power/wakeup" as many were disabled, but I still have issues as the settings do not seem to persist. When I reboot some go back to disabled. I have tried adding that to rc.conf but I do not think that is the right place, since the settings still do not persist (Sorry new to linux). I am running three monitors, two using display port and one using VGA, along with one Dragonfly DAC and a wireless KB/Mouse setup. Sometimes I get a message about "xhci not responding bad cable?" or just something along the lines of "xhci not responding". I am running a galp4 pro with the i5 processor and 64GB of ram. Let me know if there are any other suggestions.

LinkOperator:
Can we add - HP USB-C Dock G5 to the accessories... 

The G5 Dock works with dual display-port monitors (MSI Monitors https://www.newegg.com/black-msi-optix-g32cq4-32/p/N82E16824475071) Curved Screens resolutions below 1080p, full functionality. Resolutions above 1080p cause monitors to go black.
- Have connected a Logitech 1080p camera to dock USB and have completed zoom meetings on a Galago Pro.
- Dock connected USB devices have full functionality. 
- Using headphone jack for powered speakers with no issues, can select between devices plugged into the unit.
- Ethernet fully functional.
- Wake button on top of the unit will wake the machine, but not turn it on.

Let me know if you need more details, willing to test further. Love the Galago Pro and Pop!_OS, but learning to work with Linux as a choice and not disappointed.

jontis:
Comments about the Lemur Pro from system76 with this HP dock with dual monitors on display port connection, two different identical setups. This worked semi-perfect for 6 months, then the screens have started to sometimes not wake up. On both separate setups, which makes me think that the dock is not the problem. I think something has changed in the popos / lemur system. The screens are sometimes, not always detected in the settings - display tab. Even if they are detected, and the desktop expands to use them, they stay off. And sometimes, they turn on and both screens and internal display are used as expected. I tried using the idea here to "echo enabled|sudo tee /sys/bus/usb/devices/*/power/wakeup". It seemed for a while to remedy the problem, but now does not seem to work. Also, I noted that when using "echo enabled|sudo tee /sys/bus/usb/devices/*/power/wakeup", the laptop does not stand by properly, but stays on when not in use.

snrkl:
@jontis : Lemur Pro (LEMP 10) with HP HP Elite USB-C Dock G4, and I am seeing what would appear to be the same issues you are describing. 

Dock drops out reasonably often, and comes back without the monitor connected to it. I have the benefit of hundreds of these docks at work, so I have swapped all parts of the dock and cable to no avail. 

I have dmesg trace if anyone wants it...

PR #242, opened 2019-11-22
Updates to the pritner page for pop.

thomas-zimmerman:
drop reference to cononical
add package info to run hp-setup


ahoneybun:
Both articles are still very different, why is that?

ahoneybun:
I would also change Software & Updates to the Pop Shop and Repoman.

doomspork:
Bah! Sorry for the hassle. @thomas-zimmerman could you update your PR to point to `master` and not `gh-pages`? We should be good to re-open and merge then.

thomas-zimmerman:
Updated changes and made new pull request vs. master

PR #248, opened 2019-11-27
Update upgrade-pop.md

cmurray21:
Issue 746: User described discrepancy with advanced install instructions. Corrected to reflect user's experience.


ahoneybun:
What discrepancy? I also don't see that issue so I'm not sure where the number is coming from.

cmurray21:
@ahoneybun https://github.com/pop-os/pop/issues/746

PR #250, opened 2019-11-29
cudnn-10.1 is not availible

TashaSkyUp:
sudo apt install system76-cudnn-10.1 fails so suggest using 10.0

### DIFF
--- a/_articles/cuda.md
+++ b/_articles/cuda.md
@@ -25,7 +25,7 @@ sudo apt install system76-cuda-latest
 To install the cuDNN library, please run this command:
 
 ```
-sudo apt install system76-cudnn-10.1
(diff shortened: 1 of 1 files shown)


brs17:
Currently 10.1 does not work in 19.10 but does work in 18.04 so this distinction should be mentioned.

mmstick:
It should actually point to `system76-cudnn-10.2` now

ahoneybun:
@mmstick I'll fix that in a new PR though it looks like that closes this issue:

https://github.com/system76/docs/issues/266

Correct?

ahoneybun:
@mmstick I set you as the reviewer of that PR.

Issue #249, opened 2019-11-29
galp4 manual states that M.2 Interface supports PCIe gen4


mokus0:
While this is possible, the CPU does not support it so it seems unlikely. Does this system mainboard have a PCIe switch that aggregates the gen3 lanes from the cpu, or is that detail in the doc a typo?

brs17:
I do believe this is a typo that should say instead, PCIe x4. @thomas-zimmerman / @ahoneybun can confirm this.

mokus0:
I've obtained the schematic and checked it, it is indeed a typo and PCIe gen3 x4 would be the correct claim

ahoneybun:
Fixed in https://github.com/system76/docs/pull/252

PR #252, opened 2019-12-16
update specs for M.2 slot

ahoneybun:
fixes issue:

https://github.com/system76/docs/issues/249


brs17:
Shouldn't the revision history in the document itself be updated as well?

PR #255, opened 2019-12-20
add ~/

ahoneybun:
makes those commands work where ever the user is in the file system.


brs17:
Shouldn't the `mv` destination also have a leading `~/`? So that things aren't just moved to whatever directory the user is in?

PR #262, opened 2020-01-03
Add section about apt pin to Installing Ubuntu article

jacobgkau:
Users cannot install the System76 Driver on Ubuntu 19.10 until an apt preferences file has been manually added. (This is still the case, even after the kernel version change to -7625.) This PR updates the Installing Ubuntu article to include this step, so users can install the System76 Driver on their computers running Ubuntu 19.10.

### DIFF
--- a/_articles/install-ubuntu.md
+++ b/_articles/install-ubuntu.md
@@ -112,6 +112,30 @@ If you ordered a system with a discrete NVIDIA graphics card or if you added one
 sudo apt install system76-driver-nvidia
 ```
 
+#### Apt Preferences File
+
+If you are running Ubuntu 19.10 or later, you will need to manually add an apt preferences file to "pin" the System76 repository. This will tell apt to prefer System76 packages over standard Ubuntu packages. Installing the System76 Driver will not be possible until this step is completed.
+
+Create the apt preferences file here:
+
+```
+sudo gedit /etc/apt/preferences.d/system76-apt-preferences
+```
+
+Add the following six lines (seven if you count the space in the middle):
+
+```
+Package: *
+Pin: release o=LP-PPA-system76-dev-stable
+Pin-Priority: 1001
+
+Package: *
+Pin: release o=LP-PPA-system76-dev-pre-stable
+Pin-Priority: 1001
+```
+
+Save the file. Now you should be able to install the System76 Driver as described above.
+
 #### If 'nouveau.modeset=0' Was Used
 
 The `nouveau.modeset=0` modifier should be made default if your machine has NVIDIA hardware. Please run this command to edit the startup options file:


_articles/install-ubuntu.md
```diff
@@ -112,6 +112,30 @@ If you ordered a system with a discrete NVIDIA graphics card or if you added one
...
+```
+Package: *
+Pin: release o=LP-PPA-system76-dev-stable
+Pin-Priority: 1001
+
+Package: *
+Pin: release o=LP-PPA-system76-dev-pre-stable
+Pin-Priority: 1001
+```
+
```
ahoneybun: I believe we would want to run 'sudo apt update' first correct?

jacobgkau: 1. "sudo apt update" is not needed to apply apt preferences, it is only needed to refresh the package lists.
2. "sudo apt update" is already part of the instructions higher up on the page to install the driver, which is what we refer to in this section.

_articles/install-ubuntu.md
```diff
@@ -112,6 +112,30 @@ If you ordered a system with a discrete NVIDIA graphics card or if you added one
...
+```
+Package: *
+Pin: release o=LP-PPA-system76-dev-stable
+Pin-Priority: 1001
+
+Package: *
+Pin: release o=LP-PPA-system76-dev-pre-stable
+Pin-Priority: 1001
+```
```
mmstick: This could be done in one command with

```sh
echo 'Package: *
Pin: release o=LP-PPA-system76-dev-stable
Pin-Priority: 1001
Package: *
Pin: release o=LP-PPA-system76-dev-pre-stable
Pin-Priority: 1001' | sudo tee /etc/apt/preferences.d/system76-apt-preferences

Comment (jacobgkau) [reply _articles/install-ubuntu.md]: I spoke with Aaron about this earlier today. This is a neat trick, but users can get confused when running multi-line commands. I can easily see someone running the first line, not realizing they're stuck in the multi-line input because of the opening quotation mark, trying to run other commands and/or get out of the multi-line input, and potentially ending up with an invalid preferences file. It's also very possible for someone reading this article to not know what "tee" is and not understand the concept that they're inserting lines into a text file.

If we're going to insist on a CLI method for this, why wouldn't we just "sudo echo > system76-apt-preferences" instead of piping into tee?

Aaron brought up the good point that running gedit from the terminal as root wouldn't work as easily once we switch to Wayland.

For the time being, I think a GUI text editor is less likely to get typo'd (or otherwise botched) than telling someone to copy/paste lines into a text file. I've used this exact verbiage with many customers in support cases, and it seems to work well.

Comment (mmstick) [reply _articles/install-ubuntu.md]: The `tee` command is always required when redirecting an output to a file in a path owned by root, as it is not the `echo` command which is performing the redirection, but the shell. So unless the shell process is owned by root, it doesn't have permission to write.

```
$ sudo echo 'test' > /test
bash: /test: Permission denied
$ echo 'test' | sudo tee /test
test
```

mmstick: Technically `gedit admin:///etc/apt/preferences.d/system76-apt-preferences` would work with Wayland, since this requests permission to read / write.

PR #271, opened 2020-01-16
Update launcher article

jacobgkau:
Currently, we have a "Move the Launcher In 16.04+" article, which is written for Unity and is therefore outdated. We also have a very short community article that goes over adding Dash to Dock to Pop!\_OS.

This PR removes the outdated Unity article, and replaces the community Dash to Dock article with a more detailed one. This will get rid of clutter in the list of articles and make the information easier to find. Ubuntu does have an extra step needed to install the extension, which is why they are separate sections in the new version of the article.


jacobgkau:
@ahoneybun It sounds like you're saying the "Move the Launcher In 16.04+" article will be disabled? What about the "Pop! - Dash To Dock" article? Will it be disabled too? I've had more than one support case where I've linked someone to Dash to Dock and explained how to install it. The new article is more complete than the old Dash to Dock article. It also doesn't make sense to have that Dash to Dock article labelled specifically for Pop, since it's a third-party GNOME extension whether you're using Pop or Ubuntu.

How about if I add back the contents of the current/old "Move the Launcher" article under a "16.04 LTS" section at the bottom, like how we have an "Upgrading older releases" section of the upgrading article?

ahoneybun:
The Dash to Dock article won't be disabled as it's the current way of customizing the dock in Ubuntu 18.04+. I want to disable 16.04 related articles to drive the point that people should be upgraded to 18.04 if they want an LTS (and 20.04 once that is out).

jacobgkau:
If it wasn't clear based on the commit messages, I did add the old URL as a redirect to the new article, so the old link wouldn't break if someone had it hyperlinked or bookmarked.

I've added back the instructions for moving the launcher in 16.04 LTS at the bottom of the article. I think this provides a sensible update adding more thorough instructions for current Pop/Ubuntu users, while clearly demonstrating that 16.04 is deprecated since it's at the bottom. What do you think?

PR #272, opened 2020-01-17
update section about the DDI settings are they aren't in the 

ahoneybun:
Open Firmware that we ship.


brs17:
You should probably update the revision history on page 1. While you're making changes regarding the Open firmware, you may also want to consider updating page 9 which currently has a lot of incorrect information.

ahoneybun:
@jacobgkau updated!

PR #275, opened 2020-01-23
Add article about the Serval's DAC clicking

jacobgkau:
I keep having to look up these commands in Slack to send to people, and it would be better if they were in an article so people can find it themselves if they want to.

### DIFF
--- a/_articles/serval-dac.md
+++ b/_articles/serval-dac.md
@@ -17,7 +17,7 @@ section: hardware-drivers
 
 ---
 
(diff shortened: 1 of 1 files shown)


jacobgkau:
I put it under the "Hardware & Drivers" section instead of the "Graphics, Audio and Video" section because people sometimes don't know which component is clicking until we tell them. They just know it sounds like clicking, so I would expect them to look under "Hardware" first.

jacobgkau:
I considered writing "On the latest version of the Serval" or "On newer versions of the Serval," but I figured that would make the article prone to becoming outdated.

I've added the model number while keeping the wording that will make it less critical to update the article when the next Serval revision comes out.

PR #280, opened 2020-01-24
Tell people to check their path

kylerchin:
it happened to me, it'll happen to someone else.


doomspork:
Oh no! Apologizes @kylerchin and @jacobgkau for this being closed. It looks like our changes to Jekyll and switch from `gh-pages` to `master` as the target caused some hiccups.

We can reopen this PR if we retarget it to the new default branch `master`

PR #287, opened 2020-01-30
Change gh-pages to master in edit link

jacobgkau:
The "Edit on GitHub" button currently links to gh-pages, which shows a 404. This will update the edit button to point to master.


burden:
On the Jekyll site, they use the following URL, which is similar to what we've got already: https://github.com/jekyll/jekyll/edit/master/docs/{{ page.path }}

When you click, it takes you to a page where Github asks you to fork the repo to make changes. Is that something we can get into? I can create a PR, if so.

ahoneybun:
@burden that could be handy. I think it would be a good idea.

PR #290, opened 2020-01-31
add image for GNOME Tweak Tool touchpad settings

ahoneybun:
This also includes ways to install the GNOME Tweak Tool as well.


_articles/touchpad.md
```diff
@@ -21,3 +21,13 @@ Press <kbd>Fn</kbd>+<kbd>F1</kbd> to turn your laptop touchpad on/off.
 To configure your touchpad, press the Super Key <kbd><span class="fl-ubuntu"></span></kbd>, <kbd><span class="fl-pop-key"></span></kbd> then type <u>Mouse & Touchpad</u> and click on the result. The touchpad settings are under the **Touchpad** tab.
 
 ![Mouse Search](/images/touchpad/mouse.png)
+
+<u>GNOME Tweaks</u> has a few other options for the touchpad like disabling it while typing, the acceleration profile and mouse click emulation.
```
doomspork: Now that this a markdown document you don't need to use HTML tags. This can be updated to

```suggestion
_GNOME Tweaks_ has a few other options for the touchpad like disabling it while typing, the acceleration profile and mouse click emulation.
```

_articles/touchpad.md
```diff
@@ -21,3 +21,13 @@ Press <kbd>Fn</kbd>+<kbd>F1</kbd> to turn your laptop touchpad on/off.
...
+
+![GNOME Tweaks](/images/touchpad/gnome-tweak-tool.png)
+
+<u>GNOME Tweaks</u> can be installed from the <u>Pop!_Shop</u> (Pop!_OS), <u>Ubuntu Software Center</u> (Ubuntu) or from the <u>Terminal</u> with this command:
```
doomspork: This one might be an exception because the `_` in Pop really upsets markdown

```suggestion
_GNOME Tweaks_ can be installed from the _Pop!_Shop_ (Pop!_OS),_Ubuntu Software Center_(Ubuntu) or from the_Terminal_ with this command:
```


```suggestion
_GNOME Tweaks_ can be installed from the <u>Pop!_Shop</u> (Pop!_OS),_Ubuntu Software Center_(Ubuntu) or from the_Terminal_ with this command:
```

_articles/touchpad.md
doomspork: https://www.markdownguide.org/cheat-sheet/

doomspork:
Not critical _or_ urgent but these markdown docs seem to be filled with HTML markup, which somewhat defeats the purpose of using markdown. Given the new setup we could consider removing all the HTML markup and replacing it with the appropriate markdown syntax.

jacobgkau:
Nathaniel and I agree with this, Aaron is a little hesitant. We're discussing it.

burden:
> @burden would you peek this real quick? I want to ensure my advice is sound.
Underscores behave just like asterisks. Sadly, markdown does not have an option for <u></ul>

jacobgkau:
https://github.com/system76/docs - The main README does show underscores for italics.

https://github.com/system76/docs/blob/master/CONTRIBUTE.md - Aaron is concerned that the CONTRIBUTE file currently uses an HTML tag as the example, although we could update that example and then change the tags to Markdown gradually as each article gets updated.

jacobgkau:
Hang on, underscores are for italics and we want this to be underlined. That would be a visual change, not just a change in markup.

PR #282, opened 2020-01-29
rename article

ahoneybun:
@nathanielsystem76 and I believe this looks better and makes the article easier to understand.

PR #296, opened 2020-02-05
Rename windows programs article

doomspork:
Let's try #282 again but this time pointed at master :ok_hand: 

The build process should re-write `gh-pages` so we needn't worry about that.


doomspork:
You're very welcome @ahoneybun! Happy to help however I can :grin:

PR #305, opened 2020-02-12
Update Ubuntu Basics article images and text

ahoneybun:
This updates the images for 18.04 LTS and adds images for 19.10. This also adds sections for both releases.﻿


ahoneybun:
@jacobgkau I'm not against using 16:9 screenshots though our customers will load different versions of Ubuntu outside of the LTS so I think we can include them.

jacobgkau:
Our Pop articles do not have 18.04 and 19.10 screenshots. The recently-updated ones only have 19.10. You mentioned that you're updating those too? It seems redundant and cluttered to have two screenshots for every screenshot just for slightly different colors/icons, when the overall layout is identical.

PR #317, opened 2020-03-13
change version of PCI-e from 4 to 3

ahoneybun:
The current manual states that the M.2 slot is Gen 4 while it is indeed Gen 3 and this commit fixes that.


storyaddict:
Confirmed with Aaron that PCIe versioning has been changed accordingly.

PR #327, opened 2020-04-10
Link lemp9 to tech-docs.system76.com

jackpot51:


jackpot51:
That's fine, do you want to make edits to this PR?

ahoneybun:
I can make edits.

WatchMkr:
@ahoneybun is the purpose of offline for the customer to print a reference while their computer is taken apart?

ahoneybun:
Yes that would be the main reason though not a lot of people know about the printing of web pages.

WatchMkr:
Because this document will contain all models, and links to exterior sources in a more modern way, it would be better to describe in the documentation that it's printable. mdBook also formats nicely for printing.

PDFs won't be sustainable long-term.

BTW, nice work on these docs everyone. They're getting really good.

_articles/service-manuals.md
```diff
@@ -15,17 +15,17 @@ section: repairs-returns
 
 ---
 
-| Laptops      | Service Manual |
-| -------------| ---------------|
-| Adder WS     | [addw1](https://github.com/system76/docs/blob/gh-pages/service-manuals/pdfs/Adder/addw1-service-manual.pdf)
+| Laptops      | PDF | Tech Docs |
+| -------------| ----| ----------|
+| Adder WS     | [addw1](https://github.com/system76/docs/blob/gh-pages/service-manuals/pdfs/Adder/addw1-service-manual.pdf) | [addw1](https://tech-docs.system76.com/models/addw1/repairs.html)
```
jackpot51, line 20: Should be https://tech-docs.system76.com/models/addw1/README.html

jackpot51:
This is really not what was intended. The case of printing the pages is handled by printing from the website. The offline copy is incomplete in features and large in size.

We should link to the tech-docs here, alone. If a user still needs an offline copy, and is unable to get it from printing the tech-docs to a file, then I believe there is still work we need to do in tech-docs to make the PDF output better

ahoneybun:
Going to use the merge Pop Docs PR to make this change.

PR #335, opened 2020-04-17
Update images for installer

ahoneybun:
These images are for installing with 20.04 LTS.


ahoneybun:
I'm aware of the virtualbox window and we can remove that if needed.

PR #344, opened 2020-04-28
Update verbiage to differentiate launcher to dock

Socialhappiness:
Replaced in whole article except 16.04 section because it is called the launcher in that DE.
Removed non-value-add statements/words


_articles/dash-to-dock.md
```diff
@@ -42,27 +42,27 @@ On the resulting pop-up, click Install to download and install the extension.
...
 
-![Launcher on left side](/images/dash-to-dock/pop-launcher1.png)
+![Dock on left side](/images/dash-to-dock/pop-launcher1.png)
 
-To customize the launcher's settings, right-click on the "Show Applications" button and click "Dash to Dock Settings."
+To customize the dock's settings, right-click on the "Show Applications" button and click "Dash to Dock Settings."
 
 ![Dash to Dock Settings](/images/dash-to-dock/pop-launcher2.png)
 
 In the Dash to Dock Settings window, you can choose which screen edge to display the launcher on, such as the bottom of the screen:
```
ahoneybun, line 53: Replace launcher with dock.

_articles/dash-to-dock.md
```diff
@@ -42,27 +42,27 @@ On the resulting pop-up, click Install to download and install the extension.
 
 ![Enabling Dash to Dock](/images/dash-to-dock/pop-install-extension.png)
 
-At this point, Firefox can be closed. The launcher will now display on the left side of the screen, similar to Ubuntu.
```
jacobgkau, line 45: The reason I wrote "similar to Ubuntu" when I re-did this article was because I've had people say they chose Ubuntu over Pop!_OS because they don't have to open the Activities menu to see the launcher, just FYI.

_articles/dash-to-dock.md
```diff
@@ -42,27 +42,27 @@ On the resulting pop-up, click Install to download and install the extension.
...
 
-![Launcher with Panel Mode](/images/dash-to-dock/pop-launcher4.png)
+![Dock with Panel Mode](/images/dash-to-dock/pop-launcher4.png)
 
 # Dash to Dock on Ubuntu
 
-Ubuntu includes a lightweight version of Dash to Dock, which is why the launcher already displays on the left side of the screen by default. When you enable the full version of Dash to Dock, it will override the built-in version.
+Ubuntu includes a lightweight version of Dash to Dock, which is why the dock already displays on the left side of the screen by default. When you enable the full version of Dash to Dock, it will override the built-in version.
 
-To install Dash to Dock on Ubuntu, you will first need to install the "native host connector," which allows Firefox to install and remove GNOME Shell extensions. This component is installed by default in Pop!\_OS, but it is not installed by default in Ubuntu. Open a terminal and run this command to install the native host connector:
```
jacobgkau, line 65: Is there a problem with having transitional words/phrases like "first" and "now" to help move along through an article? This section reads more awkwardly after this change, IMO. "To do X, do this seemingly unrelated thing" vs. "to do X, you first need to do this prerequisite."

_articles/dash-to-dock.md
Socialhappiness: Thank you for confirming the reasoning. We can add it back.

_articles/dash-to-dock.md
Socialhappiness: The statement makes sense with your explanation, so we should leave it in

PR #358, opened 2020-04-30
Install lightworks

storyaddict:
Draft 2 complete and added photos. Ready for review.


jacobgkau:
In addition to changing the command in the terminal section from dpkg to apt (because apt will install required dependencies while dpkg will not), I also removed the explanation about how to use ls/grep and just used a glob instead. We did the glob in the DaVinci Resolve article as well, so this is consistent and simpler for a first-time install.

PR #361, opened 2020-04-30
removes the second Matlab article

ahoneybun:


jacobgkau:
Will need to merge #362 immediately after this.

PR #365, opened 2020-04-30
Update cuda.md

andtheWings:
I got error message "E: Unable to locate package system76-cudnn-10.1". Running "sudo apt search cudnn" I found that versions 10.0 and 10.2 exist, but not 10.1. Perhaps you could use the construct "latest" like you do with "system76-cuda" to make things easier.


ahoneybun:
@andtheWings thank you for the work that you have done and it looks like this was fixed in another commit so I'll close this PR. Please do continue the great work of documentation as it is always welcome!

PR #370, opened 2020-05-01
Update install-minecraft.md

kolgza:
The current page instructs users to install Minecraft by manually installing the dependencies and downloading the binary from Mojang's website. This is needlessly difficult when there are other, easier methods of installing Minecraft.

EDIT: This is my first time using git to contribute to anything, so some commits may seem…questionable. However, everything else should be fine.


kolgza:
It would potentially be a good idea to instruct the user on how to install Forge, Optifine, and also resource packs—especially since the Minecraft path might be different than usual when using this tutorial.

PR #371, opened 2020-05-02
Updating link to pop package manager doc

joemcl:


joemcl:
@btkostner you're welcome!

PR #372, opened 2020-05-03
remove reference at top to 18.04

joemcl:
I *think* the reference at the top to 18.04 is now out of date, correct? So this removes it.


joemcl:
thanks @Socialhappiness got it, so closing.

PR #376, opened 2020-05-08
Updates for new shortcut on 20.04

brunodrugowick:
This updates the markdown for this tutorial considering the new shortcut to move between workspaces on the 20.04 LTS version.

The previous shortcut was Super + Up or Down.
New shortcut is CTRL+Super+Up or Down.


ahoneybun (dismissed):
Thank you for the work that you have done! I would suggest making two sections (one for 18.04 and one for 20.04) since 18.04 is still supported. Once that is done I'll approve these changes.

jacobgkau:
When testing this, I got:

```
system76@pop-os:~$ xte 'keydown Control' 'keydown Super_L' 'key Down' 'keyup Super_L' 'keyup Control'
Unable to resolve keysym for 'Control'
Unable to resolve keysym for 'Control'
```

It needs to be `Control_L` (or `Control_R`), similar to using `Super_L` instead of just `Super`. Submitting a correction for this now.

jacobgkau:
@ahoneybun I cannot merge this since you requested changes. We currently have two supported versions that are newer than 18.04 LTS, do you still want a section in this article just for the outdated version? If so, I can add one, but it doesn't seem like it would be needed at this point and would just be more clutter to maintain, which is why I haven't added it yet.

PR #377, opened 2020-05-09
Update live-disk.md to add a table of contents

bgoodr:
There are important sections that remain hidden because there is no table of contents. Only by accident did I find those sections.


bgoodr:
Unfortunately, the GiotHub system does not show a "preview" of my suggested changes, so I cannot verify that the syntax is correct or not.

PR #386, opened 2020-05-20
Fix command to switch integrated graphics

hellofaduck:
not "intel" but "integrated"


doomspork:
Thank you for the contribution @hellofaduck! It looks like this change was addressed in another PR merged today. Please don't let this discourage you, we hope you'll continue to contribute :tada:

PR #318, opened 2020-03-13
Update graphic modes

ahoneybun:
This adds the commands for running applications though the NVIDIA GPU while in Hybrid mode.

PR #390, opened 2020-05-22
Update graphics article

ahoneybun:
There was a change that converted intel to integrated in the system76-power package. This update makes that change for the commands and other text in both articles for graphic switching.


crawfxrd:
Should be done as part of #318.

PR #392, opened 2020-05-26
Update Pop Upgrade Article

Socialhappiness:
Added workarounds currently used by support for upgrade issues, removed outdated imagery

### DIFF
--- a/_articles/upgrade-pop.md
+++ b/_articles/upgrade-pop.md
@@ -115,7 +115,7 @@ sudo sed -i 's/old-releases/us.archive/g' /etc/apt/sources.list
 sudo sed -i 's/disco/focal/g' /etc/apt/sources.list /etc/apt/sources.list.d/*.list
 ```
 
-3) Now,do the upgrade! This will have a three phases and once the download is complete, you can't change your mind. This works from a new install of 19.04, so make sure you have a backup of important data before moving forward. There may be a question about restarting services, and it is safe to answer "Yes". There may be some questions about using "maintainer" version of configuration files and using the that new version is also very likely what you want to do (and the default answer will work):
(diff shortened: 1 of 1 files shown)


_articles/upgrade-pop.md
```diff
@@ -95,28 +91,39 @@ Once restarted, the computer will be on the newly upgraded system! If you run in
...
+sudo mkdir -p /etc/apt/sources.list.d/backup
+sudo mv /etc/apt/sources.list.d/* /etc/apt/sources.list.d/backup
+sudo apt-add-repository -yn ppa:system76-dev/stable
+sudo apt-add-repository -yn ppa:system76/pop
+sudo sed -i 's/old-releases/us.archive/g' /etc/apt/sources.list
+sudo sed -i 's/disco/focal/g' /etc/apt/sources.list /etc/apt/sources.list.d/*.list
 ```
-For Pop!_OS 17.10 Run
+
+3) Now,do the upgrade! This will have a three phases and once the download is complete, you can't change your mind. This works from a new install of 19.04, so make sure you have a backup of important data before moving forward. There may be a question about restarting services, and it is safe to answer "Yes". There may be some questions about using "maintainer" version of configuration files and using the that new version is also very likely what you want to do (and the default answer will work):
```
jacobgkau, line 118: "This works from a new install of 19.04, so make sure you have a backup of important data before moving forward."
This is a little bit unclear. Maybe add a part in the middle:
"This works from a new install of 19.04, but some packages may add complications, so make sure you have a backup of important data before moving forward."

jacobgkau, line 118: "This works from a new install of 19.04, so make sure you have a backup of important data before moving forward."
This is a little bit unclear. Maybe add a part in the middle:
"This works from a new install of 19.04, but some packages may add complications, so make sure you have a backup of important data before moving forward."

PR #401, opened 2020-06-08
Changes needed for non-EFI systems

ckoul:
Pop!_OS (versions 19.10 and 20.04 for sure) on Legacy BIOS (ie non-EFI) systems will not dual-boot (by default) with Windows because of GRUB misconfiguration (in this scenario, Windows should be installed first and Pop should be installed afterwards)!

The reason users never see a grub menu is the configuration of the following lines in file: /etc/default/grub
GRUB_TIMEOUT_STYLE=hidden
GRUB_TIMEOUT=0

These two lines should be commented or removed since they do not allow grub menu to display.

Now, even if grub menu shows up, Windows will not be listed as a possible boot candidate because 'os-prober' package is not installed by default during Pop!_OS installation, so Windows operating system will not be auto-detected during initial installation of grub.

It can be installed after booting into Pop!_OS for the first time with:
sudo apt install os-prober

and after:
sudo update-grub

'Windows' boot menu entry is successfully added on the grub menu.


ckoul:
Sorry about that, I am new to this and got confused when I saw the
addition. I reinserted your section in the pull request and hope everything
is OK now.

storyaddict:
@ckoul We appreciate the revisions you've suggested, and would like to incorporate them. However, there have been some substantial revisions to this article in the time since this was opened. Please take a look at the latest version, and then create another branch, and suggest your changes again. I'm closing this PR, but look forward to working with you on the new one!

PR #404, opened 2020-06-18
Fix VM article

ahoneybun:
Pop needs at least 15GB but we recommend 20GB and this fixes a typo.

### DIFF
--- a/_articles/install-in-vm.md
+++ b/_articles/install-in-vm.md
@@ -43,7 +43,7 @@ Set the memory to 2048 MB.  The default value of 1024 MB is not enough to run Ub
 
 ![Hard Disk](/images/install-in-a-vm/create_hard_disk.png)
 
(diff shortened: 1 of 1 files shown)


_articles/install-in-vm.md
```diff
@@ -43,7 +43,7 @@ Set the memory to 2048 MB.  The default value of 1024 MB is not enough to run Ub
 
 ![Hard Disk](/images/install-in-a-vm/create_hard_disk.png)
 
-You can use the defaults here for this exercise.  Feel free to accept the defaults and clik the "Create" button.
+Pop!_OS will need at least 15GB but it is recommend to give it 20GB if possible. Feel free to accept the defaults and click the "Create" button.
```
storyaddict, line 46: Typo - should be: "but it is recommended" (not it is recommend)
Add: "After adjusting storage, feel free..."

PR #407, opened 2020-06-23
VirtualBox default resolution sometimes cannot fit installer

OtterGauze:
Adding a clarification for this should help those - such as myself - who were initially lost for solutions when installing Pop in a VM, due to the inability to change the screen resolution directly from VirtualBox without Guest Additions installed.


OtterGauze:
Any resolution other than VirtualBox's default 800x600 resolution is perfectly fine for fitting the installer and allowing access to all of the controls located all around the window.

PR #411, opened 2020-06-25
Update pop-keyboard-shortcuts.md

rcalandr:
As reported in the "notable shortcut changes", to lock the screen it's fl-pop-key + ESC, so also in the OperatingSystem section should be reported the same shortcut


ahoneybun:
Those shortcuts are for older Pop releases which would still be using Super+L to lock the screen:

Pop!_OS (up to 19.10 release) Shortcuts

rcalandr:
> Those shortcuts are for older Pop releases which would still be using Super+L to lock the screen:
> 
> Pop!_OS (up to 19.10 release) Shortcuts

Sorry, completely miss that....my fault, closing the PR

PR #416, opened 2020-07-03
Fixed href id typo

Nicekor:


ahoneybun:
It looks like this change was merged with this commit:

https://github.com/system76/docs/commit/adcfedfdbd371e6d775520e081ce131191ba7dd4

I would still like to thank you for your work and for bringing this to our attention!

PR #417, opened 2020-07-05
Update to include additional docking station compatibility

ryanolton:
My Darter Pro (darp6) is regularly connected to the CalDigit TS3 Plus. I use it to connect via DisplayPort to an external monitor (2560 x 1440), a gigabit Ethernet connection, audio (via headphone jack), as well as multiple USB connections (type A and C). I have not tried a pass-through Thunderbolt 3 connection, though the dock has one. It also supplies power to the computer (up to 87 W), which means I only have to connect one cord for everything.


ahoneybun:
@jacobgkau I think we should add a section to this article that covers the ports that don't work or haven't been tested like the pass-through Thunderbolt 3 connection on this one. Thoughts?

jacobgkau (approved):
> @jacobgkau I think we should add a section to this article that covers the ports that don't work or haven't been tested like the pass-through Thunderbolt 3 connection on this one. Thoughts?

@ahoneybun We can't fully endorse community-tested docks or verify what's been tested, that's why we have them under their own section. I added the link back to this PR like the other ones have, and added a sub-entry in the list noting that the Thunderbolt 3 passthrough hasn't been tested. I also added sub-entries on the other devices using what information was available in each of their PRs.

snrkl:
What is the process to add a community tested dock's details to the list?

PR #421, opened 2020-07-19
Update upgrade-pop.md

the-dwyer:
When I tried upgrading a couple of days ago, systemd was not a valid option for pop-upgrade.


the-dwyer:
Hi. I didn't think of the update from 19.04 because I was using 19.10.  Thanks for trying that out. I am enjoying using Pop!_OS.

PR #424, opened 2020-07-23
Replace globs with apt patterns

jacobgkau:
In apt 2.0 (released with Pop/Ubuntu 20.04), wildcard globs were disabled in apt. These four commands would not have worked since then.

At this point, [apt 2.0.3](https://lists.ubuntu.com/archives/focal-changes/2020-May/018378.html) is in Ubuntu Proposed, and it will reinstate wildcard globs when released. However, it's still better to use patterns because globs can fail due to Bash filename expansion-- if a user has a file called "nvidialogs.txt" in their current directory, then `sudo apt reinstall *nvidia*` would fail with `Unable to locate package nvidialogs.txt`. A pattern will work even if there's a filename that matches the specific pattern being used.


ahoneybun:
The messaging for that command may be a bit scary for certain customer/users. Is there a different command to do the same task?

jacobgkau:
That was not intentional, I apologize for not catching that. It [looks like](https://askubuntu.com/questions/880291/apt-get-install-with-minus-hyphen-removes-packages) a dash at the end of a package name tells apt to remove the package, even if we're using `apt install`.

The command that was in this article before used `system-config-printer-*`. When wildcard globbing is enabled (apt <1.9.6 and >2.0.3), that matches three packages:

- system-config-printer-gnome
- system-config-printer-udev
- system-config-printer-common

The first one (system-config-printer-gnome) is not installed by default in Pop!_OS or Ubuntu, so it's added when a user runs the command currently in the article. Also, there is a `system-config-printer` package that is not reinstalled with the article's current command, since the article currently has the trailing dash before the wildcard. Just wanted to point those things out.

If we wanted to keep the article's current behavior with the new apt syntax (which is what I was attempting to do with this PR-- just a syntax update, not trying to change the instructions), the new command would be:

`sudo apt reinstall ~n^system-config-printer-.*`

Or if you'd like the base `system-config-printer` package to get reinstalled as well, we could just get rid of the dash instead:

`sudo apt reinstall ~n^system-config-printer`

Thoughts on if you want to keep the old behavior or if you also want to reinstall `system-config-printer`?

ahoneybun:
I think we can update the printer commands to reinstall the packages as it should be and to avoid that warning. That would be perfectly fine.

Issue #425, opened 2020-07-29
Update System Firmware page outdated


Zooce:
https://support.system76.com/articles/system-firmware/ is outdated in the following ways (note I'm using an Adder WS -- addw2):
1. There's no Settings > Devices > Firmware. It's now Settings > Firmware.
2. The Settings > Firmware page does not have an Update button (maybe this is because my firmware was up-to-date?).
3. The `system76-firmware-cli` utility was not installed by default on my system. I had to run `sudo apt install system76-firmware` to get it.
4. Step 7 did not apply to my machine. Maybe it does apply to others?
5. The first part of Step 8 happened automatically (the BIOS was launched without me doing anything). For the second part of Step 8 (the Exit tab), there's a step missing which is to enter the Setup Utility.
6. In Step 9, there was no Load Setup Defaults, only Load Optimal Defaults. Maybe this is machine specific?

Zooce:
If someone can comment on these issues and let me know if there's anything I'm missing or need to clarify, I'm more than willing to open a PR for this.

Socialhappiness:
@Zooce Thank you for bringing this to our attention. The article has been updated.

PR #426, opened 2020-07-31
Update firmware article

jacobgkau:
Closes https://github.com/system76/docs/issues/425.

Updates the firmware article to be more clearly worded, includes pictures for easier following along, and fixes any previously outdated information.


_articles/system-firmware.md
```diff
@@ -18,52 +18,93 @@ These instructions are for System76 owners who have been prompted for a firmware
...
+
+Depending on the model, the system may reboot at this point. The system's fans may spin at full speed while the new firmware is being written. Once the firmware update process is complete, you may see a prompt to **press any key** to shut down.
+
+![Press any key to shut down](/images/system-firmware/press-any-key.jpg)
+
+After the system powers off, **press the power button** to turn it back on. On machines running Open Firmware, the system should boot normally.
+
+### Disabling the ME
+
+For machines not running Open Firmware, a few extra steps may be required after an update to ensure the Intel ME is disabled.
```
leviport, line 72: Should we change this to, "For laptops not running Open Firmware, ..."? People will probably reference this for desktops which don't have disable-able ME.

jacobgkau: For some reason, I thought "may" would alert people that section might not apply (along with the images), but you're right, it's better to specify. Changed in https://github.com/system76/docs/pull/426/commits/f6b3dc59147503131a62c808f052e17ba91c1dee.

PR #431, opened 2020-08-13
Add Lemur Pro and Dell DS1000

adeptg:
Tested and works:
- Charging (USB PD)
- Video output (via DisplayPort)
- USB type A outputs

Not tested:
- Ethernet


adeptg:
Thanks, @jacobgkau! I've updated pull request, but not sure if it's what you're asking for, sorry.

PR #436, opened 2020-08-20
Update system-firmware.md

nathandyer:
Added lemp9 to the list of devices that don't support USB-C charging during firmware updates


leviport:
Dang, I forgot about this, sorry about that. I just confirmed that lemp9 supports USB-C charging during firmware updates.

PR #441, opened 2020-08-26
Update codecs.md

nathandyer:
Updated an additional package needed in 18.10+ install instructions, and added information about the ubuntu-restricted-extras package.


jacobgkau:
Installing `libavcodec-extra` (which was already in the list) pulls in `libavcodec-extra58` as a dependency, is there a reason that needs to be added to the list explicitly?

PR #451, opened 2020-08-28
Update pop-recovery.md

Owslla:
The text at the end of line 123 says to see the instructions 'below', when it is referring to line 59 (#encrypted-disk) above.

Propose changing 'below' to 'above' at the end of line 123.


ahoneybun (approved):
LGTM thank you!

PR #456, opened 2020-09-03
Update quickstart

ahoneybun:
This addresses this feedback from a customer:

https://support.system76.com/articles/desktop-quickstart/

"Regarding the admin account: I was paying attention to the paragraph after the one you cited, which reads:
"After the encryption passphrase is set, you will proceed to set up a user account and be prompted to set the administrator password."
(Which I interpreted as two steps, with the setting being the root password.)

A person new to linux probably wouldn't be too confused by all this; it's more the expectations of someone coming from a distro with a typical old-style root account setup. An explicit line saying that root doesn't need to be set up would probably be useful for my sort of case. Thanks!"


_articles/desktop-quickstart.md
```diff
@@ -45,7 +45,7 @@ Press the circular button on the front of the machine after all components are p
...
 
-You will be prompted for the encryption passphrase first if you select that option during the account creation process. After the encryption passphrase is set, you will proceed to set up a user account and be prompted to set the administrator password. If you encrypt your hard drive, you will be prompted for the encryption passphrase every time you power on. It's important to write this password down and keep it somewhere safe. If the password is lost or forgotten, you will lose access to the contents of the drive.
+You will be prompted for the encryption passphrase first if you select that option during the account creation process. After the encryption passphrase is set, you will proceed to set up a user account and be prompted to set the administrator password. This administrator account has sudo rights but it is not the root account like some distributions have you set up like Debian. If you encrypt your hard drive, you will be prompted for the encryption passphrase every time you power on. It's important to write this password down and keep it somewhere safe. If the password is lost or forgotten, you will lose access to the contents of the drive.
```
storyaddict, line 48: If you selected the option to encrypt during the setup process, you will be prompted to set the encryption passphrase first. If you encrypt your hard drive, you will be prompted for the encryption passphrase every time you power on. It's important to write this password down and keep it somewhere safe. If the password is lost or forgotten, you will lose access to the contents of the drive.

After the encryption passphrase is set, you will proceed to set up a user account and be prompted to set the administrator password. 

**NOTE:** The administrator password and the encryption passphrase are separate. Users can set them to have the same characters, but to the OS they are different items. It is good security practice to make the encryption passphrase and the administrative password different. However, this means that when prompted for the admin password, the encryption passphrase will not work, and vice-versa.

This administrator account has sudo rights, but it is not the root account. Some distributions (like Debian) have you set up a separate root account. Ubuntu and Pop!\_OS do not.

storyaddict (changes requested):
Clarifying difference between passphrase and account types.

storyaddict (approved):
Requested changed made. Looks good.

PR #457, opened 2020-09-03
change from google doc to tech-docs for the darp6

ahoneybun:


jacobgkau (approved):
We should probably standardize whether we're linking to README.html or repairs.html, but that can be done in a future PR. Both of the links being added here are functional.

PR #458, opened 2020-09-14
Remind user to make sure install location is in path

cjriley9:
For me this was /usr/lib/cuda/bin/ but I wasn't sure if different configurations may put it in different locations so I left that out. Not sure the wording on the change is the best, but something like this would have saved me some trouble with verifying the install.


ahoneybun:
@mmstick can you review this for support?

mmstick:
The CUDA SDK paths are added to the system paths on install, so there's no need to do that. But you do have to restart to have the updated path in your environment.

ahoneybun:
It sounds like this isn't needed so I'm closing this now.

PR #460, opened 2020-09-15
Add information about transitioning to Open Firmware

jacobgkau:
Documents the option to transition to Open Firmware. This may be moved lower in the article down the road (since future products will ship with Open Firmware and won't need to transition), but for right now, this seems like a logical place to put it. The header of this section can be linked to directly in social posts, if desired.

I'm adding `apt update` to the steps for installing since an outdated version of system76-firmware-cli would silently discard the `--open` option (or any other invalid option.)


leviport (approved):

_articles/system-firmware.md
```diff
@@ -55,6 +56,14 @@ sudo system76-firmware-cli schedule
 
 Reboot the system using `sudo systemctl reboot` to proceed with the update, or use `sudo system76-firmware-cli unschedule` to cancel the update.
 
+### Transitioning to Open Firmware
```
jackpot51, line 59: Steps should be added for transitioning back

leviport: We'll also need to include expected feature differences between the two so they know what to expect before transitioning.

jacobgkau: Added the command to transition back and some expected differences in https://github.com/system76/docs/pull/460/commits/eae2c0032bab391290a45c9394caac5774171b29, let me know if either of you think anything important is missing or if anything in the list of differences looks inaccurate.

jackpot51 (approved):

leviport (approved):

PR #461, opened 2020-09-16
Removed error in the Instructions

bytenaija:
Removed error in the system-config-printer re-installation instructions


_articles/add-a-printer-pop.md
```diff
@@ -53,7 +53,7 @@ To look at the CUPS (Common Unix Printing System) configuration and status windo
 
 [localhost:631](http://localhost:631)
 
-The status window will show current print jobs, detected printers, and other information about the printing system. If you would like to share this printer with others on your local network, click on the 'Admin' link, under Server, click on the "Share printers connected to this system" and save the changes. Other computers on your network should than see that printer. When there is a prompt for your username and password, use your user name, and password used to login.
+The status window will show current sudo apt install --reinstall ~n^system-config-printerprint jobs, detected printers, and other information about the printing system. If you would like to share this printer with others on your local network, click on the 'Admin' link, under Server, click on the "Share printers connected to this system" and save the changes. Other computers on your network should than see that printer. When there is a prompt for your username and password, use your user name, and password used to login.
```
jacobgkau, line 56: Why is the command pasted in the middle of a sentence?

_articles/add-a-printer-pop.md
```diff
@@ -64,7 +64,7 @@ sudo apt install --reinstall cups cups-client
 This will reinstall <u>CUPS</u>, the main printing software, which can help with generic issues.
 
 ```
-sudo apt install --reinstall ~n^system-config-printer
+sudo apt install --reinstall system-config-printer
```
jacobgkau, line 67: This was not an error, `~n^system-config-printer` will install any package with `system-config-printer` in the name. This includes:

```
system-config-printer
system-config-printer-gnome
system-config-printer-udev
system-config-printer-common
```

The revised command would only install `system-config-printer`. Is there a reason you would not recommend reinstalling the other three packages?

jacobgkau (changes requested):

bytenaija:
I actually get this error when I try that command
```no matches found: ~n^system-config-printer```

File: _articles/add-a-printer-pop.md (hunk 2 above)
Comment (bytenaija) [reply _articles/add-a-printer-pop.md:67]: I actually get this error when I try that command
```no matches found: ~n^system-config-printer```

**Comment by jacobgkau** [comment]
> I actually get this error when I try that command ```no matches found: ~n^system-config-printer```

I have confirmed the command still works properly on a clean Pop!_OS 20.04 machine. What version of Pop are you running? Please provide the complete output of these commands:

```
cat /etc/os-release
sudo apt update
apt policy apt
sudo apt install --reinstall ~n^system-config-printer
```

jacobgkau:
Closing this PR for now, as the listed command works on 20.04 LTS and 20.10, and we haven't heard a response from @bytenaija to suggest why it may not be working for them. I'd encourage anyone having issues with this command to open an issue or reply to this PR with more details so we can determine the cause.

PR #466, opened 2020-09-21
Separate transition to open firmware and firmware update articles

jackpot51:
I also escaped some instances of Pop!_OS that were not markdown escaped. Some parsers care, others don't.

### DIFF
--- a/_articles/system-firmware.md
+++ b/_articles/system-firmware.md
@@ -137,4 +137,4 @@ If you’re receiving the firmware update notification but there are no updates 
 
 ### Switching Between Proprietary Firmware and System76 Open Firmware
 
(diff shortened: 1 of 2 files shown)


thomas-zimmerman (approved):
Clear and concise on the new article. We may want to expand a bit on transitioning between system76 open firmware and proprietary (ie, list the model that have the option--there won't ever be many).

jackpot51:
> Clear and concise on the new article. We may want to expand a bit on transitioning between system76 open firmware and proprietary (ie, list the model that have the option--there won't ever be many).

The models are listed at the top of the article. Adder WS (addw2) and Gazelle (gaze15). Is that what you wanted?

jacobgkau (approved):
Looks good to me at this point. No need to copy the images from the updating article since we're linking back to that article for the second half of the process anyway.

PR #471, opened 2020-10-07
Add note about galp keyboard behavior on Open Firmware

crawfxrd:
Ref: system76/ec#123


jackpot51 (approved):

jacobgkau:
We haven't published Open EC for galp4, and this would only apply with Open EC, correct? (This machine already shipped with "Open Firmware," so maybe the note should refer to Open EC instead?)

storyaddict (approved):
No issues or typos found.

PR #473, opened 2020-10-12
Removed appfolders extension

VentGrey:
GNOME 3.36 already supports app folder creation and management, GNOME 3.38 also supports appfolder re-ordering


ahoneybun:
I think it would be better to add a tag below replacing the 19.10 one or adding onto it since some users will stay on 18.04 LTS where this is needed.

VentGrey:
Added the missing tag :sweat_smile: I totally forgot about 18.04

jacobgkau (approved):

PR #479, opened 2020-10-28
Update to 20.10 instructions

Socialhappiness:
changed to 20.10 and modified the older releases upgrade section


storyaddict (approved):
Fixed a typo.

PR #482, opened 2020-10-29
Update pop-20.10-upgrade-help.md

linuxgnuru:
added decryption of NVME option as well as initial mount which will work for non-decypted drives


storyaddict (changes requested):
May want to add the "data-root" note from Recovery article, but the current changes are needed and make sense.

storyaddict (approved):
Adding clarification on decryption and disk mounting.

ahoneybun (approved):
Looks even better now!

PR #485, opened 2020-10-30
update panel info from 144Hz to 60Hz

ahoneybun:
Updates the panel info from panelook.


storyaddict (approved):
Re-uploading corrected PDF file.

PR #489, opened 2020-11-04
fix package manager link

ahoneybun:


storyaddict (approved):
fixed incorrect hyperlink

PR #495, opened 2020-11-19
Update wireless.md

storyaddict:
Changing systemctl restart command from "network-manager" to the correct "NetworkManager"


ahoneybun (approved):
LGTM!

PR #455, opened 2020-09-03
Switch macos to popos

storyaddict:
Changes have been made. Ready to merge after final review.

PR #499, opened 2020-12-01
Update Pop!_OS 20.04 release notes

foo-dogsquared:
I've noticed there is a duplicate of the previous section so I removed it. I also introduced some consistency with the list.


jacobgkau (approved):
Looks like the duplicated list may have been introduced in this unrelated PR: https://github.com/system76/docs/pull/455 

Thank you for the correction.

PR #501, opened 2020-12-07
Update dual-booting.md on its very first paragraph (to use recovery m…

kirameister:
…ode in order to properly use GParted onto the Pop!_OS partition)


anneris (commented):
It could be beneficial to mention that using the Pop!\_OS live environment is also an option.

`To dual boot Pop!\_OS alongside another OS install Pop!\_OS first. Upon completing the installation of Pop!\_OS, reboot your computer in Recovery mode by holding down the spacebar during boot, which allows you to select Recovery mode in the systemd-boot list.  Alternatively you can use your Installer USB Key to boot into a Pop!\_OS Live Environment.  Then once booted into Pop!\_OS recovery mode or Live Environment, use GParted to resize the root partition (the largest partition) to make room for Windows 10 (35GB is the minimum required).  `

kirameister:
@anneris : Thank you for your review. I've updated the paragraph as suggested.

jacobgkau (approved):
Confirmed GParted isn't able to resize a root partition while booted into that root partition, so rebooting into a live environment would be necessary. Thank you for the contribution.

PR #502, opened 2020-12-17
Update service-manuals.md

linuxgnuru:
Added lemp10 link


storyaddict (approved):
Looks good. Just adding the manual link to the table.

leviport:
Just a thought: do you want to link to the [repairs page](https://tech-docs.system76.com/models/lemp10/repairs.html) rather than the [README page](https://tech-docs.system76.com/models/lemp10/README.html)? I'm sure it's no big deal either way, but the old links do go directly to the service manual PDFs, so it seems a bit more consistent. I saw lemp9 also links to the README, so other models would probably have to be updated as well.

PR #506, opened 2020-12-21
add thelio-mega and galp5

ahoneybun:


jacobgkau (approved):
Both new links are present and lead to the intended location (repairs page for each respective model.)

image

PR #508, opened 2020-12-24
update version numbers

ahoneybun:


storyaddict (approved):
Changes look good, and it makes sense to update version numbers and separate 20.04 and 20.10 information.

PR #510, opened 2020-12-24
add specific version for cuDNN

johndpope:
will need to dig more into other version (ongoingly)


ahoneybun:
@mmstick can you review this PR for support?

mmstick:
Version changes each time we update the repo

PR #512, opened 2020-12-24
replace commands

ahoneybun:
This affects any articles where we have the following command:
sudo cp /etc/resolv.conf /mnt/etc/

If we change it to:
sudo cp -n /etc/resolv.conf /mnt/etc/

then customers won't get an error message and possibly freak out

I've found this in the following but might be in more areas.
https://support.system76.com/articles/bootloader/
https://support.system76.com/articles/pop-recovery/


thomas-zimmerman (approved):
LGTM

PR #515, opened 2020-12-26
Remove backups apps without linux native support

andirsun:
Box.com doesnt have a native client for linux. 
I think you can drop the reference for this app.


jacobgkau (approved):
Thank you for the contribution.

- While you removed one reference to box.com, another one was remaining. (It looks like the links were mistakenly duplicated in this unrelated commit: https://github.com/system76/docs/commit/27590095d3a2dd54b03413eb9feb9fb170194019)
- box.com does have an official CLI client that works on Linux (it can be compiled from source or installed via npm.) I've confirmed that client installs and runs.

This article also needed some grammatical and formatting revisions, so I've done that, removed the duplicate links, and marked the box.com link as "CLI only" to distinguish it from the two options listed that offer GUI clients. (I also moved box.com to the end of the list since their GUI client doesn't support Linux.)

Knowing that the CLI client and web interface do work on Linux, I think the original authors would still want to have it listed since that section refers to options for free storage for offsite backups.

PR #516, opened 2020-12-27
Add note for executing shell script

jackblk:
Adding a note for executing shell script via custom shortcut.

Since other distro like Linux Mint has an option to browse to the script and handle by itself, in PopOS we only have command, so a note will help future users.

### DIFF
--- a/_articles/keyboard-shortcuts.md
+++ b/_articles/keyboard-shortcuts.md
@@ -42,7 +42,7 @@ Enter the information about your shortcut, including what the name will be, the 
 
 Then click "Add", and your shortcut will be active.
 
-*Note*: For executing shell script, use absolute path instead of relative path. Example: `sh /home/username/Documents/script.sh` instead of `sh ~/Documents/script.sh`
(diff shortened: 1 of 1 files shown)


_articles/keyboard-shortcuts.md
```diff
@@ -42,6 +42,8 @@ Enter the information about your shortcut, including what the name will be, the
 
 Then click "Add", and your shortcut will be active.
 
+*Note*: For executing shell script, use absolute path instead of relative path. Example: `sh /home/username/Documents/script.sh` instead of `sh ~/Documents/script.sh`
```
leviport, line 45: Thanks for the contribution! This sentence reads a little strangely to me. Do you think "In order to execute a shell script, use its absolute path instead of its relative path" would sound better?

leviport (changes requested):

_articles/keyboard-shortcuts.md
jackblk: Ah yes, it will be better that way 👍.

leviport (approved):
Looks great, thanks again!

PR #518, opened 2020-12-28
Add Dell WD19 and Plugable ud-ultcdl dock info

skylord123:
I use the Dell WD19 daily at work and the Plugable ud-ultcdl at home with great success. I'm using this with a brand new Gazelle (gaze15) laptop. Note: Just as it states on System76's website to use docks with this laptop you must have the GTX 1660 Ti.

I run the WD19 with three monitors plus the built in laptop display for a total of 4 displays. I so far have only tested the Plugable at home with a single monitor.

Power delivery does NOT work for either of these docks for the Gaze15 (I don't think the Gazelle can even charge via type-c but I may be wrong). My solution is just to leave an extra triple prong power cable for my laptop's power brick plugged in and ready to go next to where my laptop sits on my desk.

There are a couple issues with the WD19 (nothing a show stopper though):
- After you lock the computer and displays go to sleep sometimes one, two, or all three screens (usually it's just one but occasionally more) will fail to turn back on or will have really bad artifacts showing on screen. Unplug and then re-insert the cable back into the machine and everything goes back to normal (this isn't every time and happens about 5~10% of the time). Re-arranging the displays within the OS can also have this same problem.
- I have to leave the dock unplugged until I decrypt and fully login to the device. If the dock is plugged in when I try to decrypt the drive it will hang and unplugging the dock does nothing to fix this. You have to power cycle the laptop with the dock unplugged to decrypt. Same problem with first login after decrypt (logging out and in works fine though which is strange). This problem didn't pop up until I installed the displaylink drivers (but I need the displaylink drivers for the Plugable dock).

So far I haven't ran into any issues with the Plugable UD-ULTCDL but if I do I will make sure to update this thread.

### DIFF
--- a/_articles/use-docking-station.md
+++ b/_articles/use-docking-station.md
@@ -70,10 +70,10 @@ Community members have reported that the following docks work with our products:
  - [Dell DS1000](https://www.dell.com/support/manuals/us/en/04/dell-dockstand-ds1000/ds1000_docking_stand_ug_publication/technical-specifications?guid=guid-1ad58fe1-dd33-4ebc-bac1-8e6a9083eb35&lang=en-us) [[community-tested](https://github.com/system76/docs/pull/431) on an Intel system]
    - Ethernet port not tested.
  - [Dell WD19TB Thunderbolt Dock](https://www.dell.com/en-us/work/shop/dell-thunderbolt-dock-wd19tb/apd/210-arik/pc-accessories) [[community-tested](https://github.com/system76/docs/pull/206) on an Intel system] <sup>1</sup>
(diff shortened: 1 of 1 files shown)


jacobgkau:
Thank you for the contribution! I made the following changes:
- Corrected the name of the Dell dock to `Dell WD19 Dock` (this dock is not a Thunderbolt dock, unlike the WD19TB)
- Added note about the screen wake issue on the Dell dock.
- Removed the <sup>1</sup> from the Pluggable dock, since you said it does need the DisplayLink driver.
- Moved the W19 above the W19TB since the end of a word would generally come before a character in alphabetical order.
- Changed from `community-tested on an Intel system` to `NVIDIA system` since that's what we consider the Gazelle.

> Power delivery does NOT work for either of these docks for the Gaze15 (I don't think the Gazelle can even charge via type-c but I may be wrong).

You are correct, the gaze15 does not support USB-C charging, so that doesn't mean anything is wrong with the dock's charging capabilities for the systems that would support it.

jacobgkau (approved):

PR #519, opened 2020-12-28
Update battery.md

linuxgnuru:
Comment out portions relating to charging thresholds and open firmware options that are still WIP


storyaddict (approved):
Commenting out WIP section that was causing user-confusion.

PR #520, opened 2020-12-29
Update dual-booting.md

anneris:
Added extra options for partitioning in the Windows setup, to better correspond with the screenshots.


jacobgkau (approved):
Reviewing this, the old sentence didn't actually make sense (`selected new press the 'Apply' button` skipped some words and steps.) This article needs some more work, but this is an improvement for now. Thank you for the contribution!

PR #526, opened 2021-01-13
Update service-manuals.md

linuxgnuru:
Adding tech doc for Adder WS (addw2)


storyaddict (approved):
Adding link to Service manual page table.

PR #533, opened 2021-01-28
Remove stray slash from list of commands

jacobgkau:
I was copying commands out of this article this morning and noticed there's a `/` in the middle of two command lists. It looks like these were added in https://github.com/system76/docs/pull/512 by mistake (as `/` is not a valid command, it was added to the Pop article but not the Ubuntu article, and that PR didn't say anything about why these lines were added.) The rest of the commands work normally without doing anything at this step.


leviport (approved):
Whoops. Nice catch

PR #535, opened 2021-02-01
updated to reflect changes in 20.10

markpendlebury:
Updating keyboard-shortcuts.md to better describe the process of modifying and updating keyboard shortcuts for 20.10


jacobgkau (approved):
Thank you for the contribution. This article definitely needed to be updated (the update isn't just for 20.10-- the new keyboard shortcuts interface was backported to Pop 20.04 as well.) 

There were a number of grammatical issues (the most noticeable being missing punctuation, but also headings written as sentences and `verticle` used in place of `vertical`), and the overall structure was changed more than it needed to be for an update like this. Also, some of the screenshots started with customizations already applied.

I've pushed a commit to rectify these things (bringing the structure more in line with how it started), and will now be merging this.

PR #537, opened 2021-02-03
howto on re-enabling raw image previews in nautilus.

bflanagin:
This how-to is based on the work done in this issue [pop #1484](https://github.com/pop-os/pop/issues/1484) adding raw image thumbnails to work in 20.04 and 20.10.

### DIFF
--- a/_articles/fix-raw-image-previews.md
+++ b/_articles/fix-raw-image-previews.md
@@ -14,9 +14,9 @@ section: pop
 
 ---
 
-## Install Rawtherapee
+## Install RawTherapee
 
-Raw Therapee can handle a large array of RAW image formats. We will be using rawtherapee's ability export raw images into pngs for this howto.
+RawTherapee can handle a large array of RAW image formats. We will be using RawTherapee's ability export raw images into pngs for this howto.
 
 ```
 sudo apt install rawtherapee
--- a/_articles/fix-raw-image-previews.md
+++ b/_articles/fix-raw-image-previews.md
@@ -39,15 +39,11 @@ Save the text file as `rawtherapee.thumbnailer`. We will move this new file into
 
 ## Moving to thumbnailer folder
 
-As with most files outside of your home directory you will need elevated privilages to move this file to the proper location.
+As with most files outside of your home directory, you will need elevated privilages to move this file to the proper location. Run the following command to move the thumbnailer file into place (replacing `<pathtofile>` with the actual path):
 
 ```
 sudo mv <pathtofile>/rawtherapee.thumbnailer /usr/share/thumbnailers
 ```
-
- Be sure to change <pathtofile> to the actual path.
-
-
 
 ## Clearing previous thumbnails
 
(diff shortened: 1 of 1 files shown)


jacobgkau:
@bflanagin I've made a number of grammar and formatting fixes that you can view in the above commits. My only other question: would you be opposed to providing a command for creating the thumbnailer file in the proper location, instead of saving it and then moving it into place? That's how I would typically handle this in a support case, for the following reasons:

- It doesn't require the user to know what to put for `\<pathtofile>'. (Users new to Linux or not interested in the filesystem layout may have trouble with this.)
- It doesn't require the user to know the difference between a text editor and a document editor. (Someone who doesn't want to use `gedit` will likely know they can use something else, while someone who doesn't know about `gedit` might not know they need to use it instead of something like LibreOffice Writer.)

For example, we could say something like this:

---

Nautilus uses thumbnailer scripts to create thumbnails for images. Create a thumbnailer for raw images using this command:

```
gedit admin:///usr/share/thumbnailers/rawtherapee.thumbnailer
```

Copy and paste the following text into the file:

```
[Thumbnailer Entry]
TryExec=/usr/bin/rawtherapee-cli
Exec=/usr/bin/rawtherapee-cli -s -n -Y -f -o %o -c %i 
MimeType=image/x-arw;image/x-bay;image/x-canon-cr2;image/x-canon-crw;image/x-cap;image/x-cr2;image/x-crw;image/x-dcr;image/x-dcraw;image/x-dcs;image/x-dng;image/x-drf;image/x-eip;image/x-erf;image/x-fff;image/x-fuji-raf;image/x-iiq;image/x-k25;image/x-kdc;image/x-mef;image/x-minolta-mrw;image/x-mos;image/x-mrw;image/x-nef;image/x-nikon-nef;image/x-nrw;image/x-olympus-orf;image/x-orf;image/x-panasonic-raw;image/x-panasonic-raw2;image/x-pef;image/x-pentax-pef;image/x-ptx;image/x-pxn;image/x-r3d;image/x-raf;image/x-raw;image/x-rw2;image/x-rwl;image/x-rwz;image/x-samsung-srw;image/x-sigma-x3f;image/x-sony-arw;image/x-sony-sr2;image/x-sony-srf;image/x-sr2;image/x-srf;image/x-x3f;image/x-adobe-dng;image/x-portable-pixmap;image/tiff;
```

Then save the file and close Gedit.

---

(Realistically, I would normally use `sudo gedit` instead of `admin://`, but the above command will work in Wayland where `sudo gedit` will not.) This would also make the instructions significantly shorter by eliminating a section.

jacobgkau:
Well, while running through this to make sure it works, I found that `sudo gedit /path/to/new/file` creates a new file like I expected:

Screenshot from 2021-02-04 14-13-06

But `gedit admin:///path/to/new/file` does not create a new file:

Screenshot from 2021-02-04 14-13-16

That's an inconsistency that I wasn't aware of. However, I also confirmed that `sudo gedit` actually does work in Wayland:

Screenshot from 2021-02-04 14-16-51

So it looks like by trying to consider Wayland, I overcomplicated things. I've changed the command to `sudo gedit`. (If this were to stop working in the future, we could just add a `touch` command before the `gedit admin://` command.)

Looks good now and confirmed that it works.

jacobgkau (approved):

PR #538, opened 2021-02-10
New threshold article

ahoneybun:
This adds a Threshold article to examine the new feature and it's current limitations for Open Firmware and Open EC systems.


_articles/battery.md
```diff
@@ -57,19 +57,6 @@ The default values of 0 for the start threshold and 100 for the end threshold
 disable the feature.
 -->
```
crawfxrd, line 58: This  commented out block isn't needed since the info is being moved.

ahoneybun: So I should remove that sentence completely?

crawfxrd: The whole thing. It's copied verbatim to the new article.

ahoneybun: New article? This is the new article.

jacobgkau:
@ahoneybun The current format of the article is:
```
## Some additional notes
### FlexiCharger (proprietary firmware)
### Charging Thresholds
#### Open Firmware
```
The additional note about the feature not being complete is only relevant to open firmware, and the explanation of charging thresholds is relevant to both open firmware and FlexiCharger. I'm going to try reorganizing the article like this instead:
```
Introduction
## Charging Thresholds (open firmware)
### Some additional notes
## FlexiCharger (proprietary firmware)
```

jacobgkau:
> The larger the spread between the start and end charge percentages, the longer the physical battery will last. For optimal longevity, charge cycles of 40%~80% are a good idea.

Our "maximum lifespan" profile uses a range of 50%-60%. (The profiles were defined after research into thresholds and the current presets that other manufacturers use.) I'm going to remove/modify some of this info so as not to conflict with the presets and confuse people.

_articles/battery.md
crawfxrd: No, `battery.md` is the existing article. `laptop-battery-thresholds.md` is the new article.

jacobgkau: I've taken care of this in https://github.com/system76/docs/pull/538/commits/7040adff30dd57c2b4c6fe1537a6671373689174.

jacobgkau (approved):
@ahoneybun Please make sure you're okay with how this reads now. If you are, then it should be good once https://github.com/system76/docs/pull/539 has been merged.

When the GUI is implemented, it should be easy to add an additional "Using the GUI:" section above the current two sections.

ahoneybun:
Looks great @jacobgkau I have merged #539 so I will merge this one now.

PR #539, opened 2021-02-10
Open firmware system list

ahoneybun:
This adds an open firmware system list to the articles and will be used in the battery thresholds article.

### DIFF
--- a/_articles/open-firmware-systems.md
+++ b/_articles/open-firmware-systems.md
@@ -33,45 +33,45 @@ Here is a [link](https://github.com/system76/ec) to the source code for our Open
 
 <table>
 <tbody>
-<tr style="height: 23px;">
-<td style="height: 23px;">System </td>
-<td style="height: 23px;">Open Firmware</td>
-<td style="height: 23px;">Open EC</td>
+<tr>
+<td>System </td>
+<td>Open Firmware</td>
+<td>Open EC</td>
 </tr>
-<tr style="height: 23px;">
-<td style="height: 23px;">galp5</td>
-<td style="height: 23px;">Yes</td>
-<td style="height: 23px;">Yes</td>
+<tr>
+<td>galp5</td>
+<td>Yes</td>
+<td>Yes</td>
 </tr>
-<tr style="height: 23px;">
-<td style="height: 23px;">galp4</td>
-<td style="height: 23px;">Yes</td>
-<td style="height: 23px;">No</td>
+<tr>
+<td>galp4</td>
+<td>Yes</td>
+<td>No</td>
 </tr>
-<tr style="height: 23px;">
-<td style="height: 23px;">lemp10</td>
-<td style="height: 23px;">Yes</td>
-<td style="height: 23px;">Yes</td>
+<tr>
+<td>lemp10</td>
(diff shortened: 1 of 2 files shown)


_articles/open-firmware-systems.md
```diff
@@ -0,0 +1,77 @@
...
+
+- Keyboard
+- Touchpad 
+- Power Button.
+
+Here is a [link](https://github.com/system76/ec) to the source code for our Open EC.
+
+<table>
+<tbody>
+<tr style="height: 23px;">
```
jackpot51, line 36: Is this height style necessary? Would it look okay without it?

ahoneybun: Looks like it's not. I removed it in the newest commit.

jackpot51: Awesome, easier to read it now

_articles/open-firmware-systems.md
```diff
@@ -0,0 +1,77 @@
...
+<td>Yes</td>
+<td>No</td>
+</tr>
+<tr>
+<td>lemp10</td>
+<td>Yes</td>
+<td>Yes</td>
+</tr>
+<tr>
+<td>lemp9</td>
```
jackpot51, line 57: What order are these being sorted by? I would normally prefer alphabetical order

ahoneybun: I don't believe there is a real order though I did put the newer versions at the top. I'll correct that in the next commit.

ahoneybun: New commit has the table list updated and in markdown instead of HTML.

jackpot51 (commented):
addw2 has been shipping with open firmware and open EC for a while. Also oryp6.

_articles/open-firmware-systems.md
```diff
@@ -33,10 +33,12 @@ Here is a [link](https://github.com/system76/ec) to the source code for our Open
 
 | System | Open Firmware | Open EC |
 |:------:|:-------------:|:-------:| 
+| addw2  | Yes           |         |
```
jackpot51: addw2 has Open EC

jackpot51:
It might also be worth linking to http://support.system76.com/articles/transition-firmware/. addw2 and darp6 shipped with different firmware than what is now available, so users could manually transition to the open firmware and open EC firmware manually

_articles/open-firmware-systems.md
ahoneybun: I fixed that in this commit: https://github.com/system76/docs/pull/539/commits/0bf9efbbd6a5ebda153ce5e83332f62d901abde6

ahoneybun:
> It might also be worth linking to http://support.system76.com/articles/transition-firmware/. addw2 and darp6 shipped with different firmware than what is now available, so users could manually transition to the open firmware and open EC firmware manually

Added that in commit https://github.com/system76/docs/pull/539/commits/da70aa76021b3a339a744077e5f0c7b77c820748 .

jackpot51 (approved):
This looks great to me now!

jacobgkau (approved):

PR #541, opened 2021-02-11
update image and steps

ahoneybun:
This is how the article section looks like now. 

ksnip_20210211-095259


jacobgkau:
The old `woeusb.png` file is not referenced in any other articles. Is there a reason you're adding a new image and leaving the old one around instead of just replacing it?

linuxgnuru (approved):

jacobgkau:
In addition to my above comment that was not addressed, I also see that the filesystem type is labeled "2" in the photo, but listed under number 1 in the description. I will work on a PR to address this and several other issues currently open about this article.

PR #542, opened 2021-02-11
audio: Organize config tweaks, add a config tweak, improve first steps.

### LINKED ISSUE #491: Workaround for audio crackling in 20.10 with Thunderbolt dock
OS Version: Pop_OS 20.10

This might be something worth mentioning in the "Fix Audio Issues" docs.

I'm running my Darter Pro (darp5) with an WD19TB a Thunderbolt docking station and I had no issues with it but a crackling when playing audio through my amplifier. This problem also only occurred when the amplifier was connected to the dock. Connected directly to the laptop via USB everything was fine.

A bit of research later and I found out this is due to the energy saving in pulseaudio. Doing the following helped:

Comment out the following line in `/etc/pulse/default.pa`

```
load-module module-suspend-on-idle
```
and restart pulseaudio `systemctl restart --user pulseaudio`

Thanks to [Tim](https://unix.stackexchange.com/questions/565886/how-to-disable-the-power-saving-for-snd-hda-codec-realtek) on StackExchange

jacobgkau:
Closes #491.

Makes the following changes:
- Fixes the section added in https://github.com/system76/docs/pull/322 to explain what it's doing, not be specific to an OS version that is now obsolete, and not have a broken link (formatted as an image when it shouldn't have been) at the end.
- Moves the above section into a new "Configuration Tweaks" section, and adds a solution suggested in https://github.com/system76/docs/issues/491.
  - Instead of the suggested wording, used the exact commands from the existing [Serval DAC article](https://support.system76.com/articles/serval-dac/) added in https://github.com/system76/docs/pull/275 for consistency and ease of use.
  - This new section comes first in the list of tweaks, because the tweak from https://github.com/system76/docs/pull/322 was added in response to a specific bug that has reportedly been fixed in Ubuntu (but could always come back later.)
- Fixed some rough wording (`reseting` isn't a word, and restarting =/= resetting) that was introduced in https://github.com/system76/docs/pull/420.


ahoneybun (approved):
LGTM! Thank you!

PR #543, opened 2021-02-12
Update dual boot drivers

ahoneybun:
This adds links to the drivers and reorders the article as well.

### DIFF
--- a/_articles/dual-booting.md
+++ b/_articles/dual-booting.md
@@ -13,6 +13,18 @@ section: pop
 
 ---
 
+### Step 1: Create Install Media
+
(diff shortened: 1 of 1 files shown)


storyaddict (changes requested):
1. Please move "using another drive" to the first position and indicate that it is the preferred method.

2. Please add a note, about Windows Device Manager showing some devices as missing drivers. And that is normal, and can be disregarded.

ahoneybun:
NOTE: The Windows Device Manager may show that some devices are missing drivers, that is normal, and can be disregarded.

storyaddict (approved):

PR #544, opened 2021-02-13
Adds section and link to download the OS image

CrustyBarnacle:
Adds section and link to download the OS image, as is expected bases on the opening paragraph "
The following guide describes how to download the Pop!\_OS.iso image, write it to a flash drive, and install it on the hardware of your choice."


ahoneybun:
Thank you for your PR! We do have a link to the live disk article which goes over downloading the OS image already.

http://support.system76.com/articles/live-disk/

CrustyBarnacle:
OK, that was not clear, but I do see the link in the linked-to article. Still think the section/article title could be improved to _Download Image and Make Bootable Drive_, or such.

ahoneybun:
@storyaddict what are your thoughts on this?

storyaddict (approved):
Good quality of life addition. Thanks @CrustyBarnacle

PR #545, opened 2021-02-15
Fix typo "Text Editer" -> "Text Editor"

ehudon:
Hi System76 folks,

I spotted a very small typo while reading this page: https://support.system76.com/articles/switch-from-macos-to-popos/

I thought I would fix it while I'm at it. 😃


ahoneybun:
Thank you for that fixed and the PR!

ahoneybun (approved):

Issue #396, opened 2020-06-01
Desktop Issue


akashpapnai:
Could not go to Desktop while anywhere on the screen.
`Super + D` doesn't work.

leviport:
The Super + D shortcut is no longer a default in Pop!_OS. Can you let us know where you see it mentioned in documentation so we can update that?

Thanks!

akashpapnai:
It's here:
https://support.system76.com/articles/pop-keyboard-shortcuts/
Search for " Show Desktop ".
And can you tell me the new shortcut for the same.
Thanks in advance.

leviport:
Thanks, we'll get that updated! 

It wasn't replaced by anything, so if you want that shortcut back, you'll have to set it manually in the "keyboard shortcuts" section of Settings.

akashpapnai:
Ok, Thanks.
It took sometime for me Search the correct word.
It's " Hide all normal window " just in case someone else read this.

PR #547, opened 2021-02-16
Remove obsolete keyboard shortcuts

### LINKED ISSUE #454: "Windows" key + L is still listed as screen lock

jacobgkau:
Closes #454.

As the current supported releases of Pop!_OS are 20.04 LTS and 20.10, and the current keyboard shortcut scheme is no longer new, the notes and entire sections referring to older versions of Pop are no longer needed (and have started creating confusion, in at least #454 and https://github.com/system76/docs/issues/396.) This PR removes all shortcuts that no longer work in the current releases, and moves the remaining shortcuts into more appropriate categories where possible.

### DIFF
--- a/_articles/pop-keyboard-shortcuts.md
+++ b/_articles/pop-keyboard-shortcuts.md
@@ -50,14 +50,14 @@ Directional actions can use either the standard arrow keys or their Vim equivale
 
 #### Manipulate windows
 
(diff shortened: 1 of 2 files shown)


ahoneybun (approved):
LGTM and these are great improvements as well!

leviport (approved):

PR #548, opened 2021-02-17
Small suggested changes in upgrading article

hobbsy:
Is there a small error in the update article?

The folder created using mkdir in step 2 for backups was named /etc/apt/backup

In step 4 it refers to a folder named /etc/apt/sources.list.d/backup


ahoneybun (approved):
LGTM Thank you for the PR! :tada:

PR #549, opened 2021-02-19
add images and new article

ahoneybun:


storyaddict (changes requested):
Great start. Let's also add Terminal/TTY commands that will accomplish the same things.

storyaddict (changes requested):
Let's make the new user an admin instead of regular user.

storyaddict (approved):
Looks good! Thanks for making these changes.

PR #550, opened 2021-02-20
Update dual boot article

### LINKED ISSUE #319: Documentation for dual-booting windows does not cover full-disk-encryption
My Darter Pro came from the factory with full-disk-encryption, and the docs for dual-booting Windows don't work. Specifically, I cannot shrink the partition.

https://support.system76.com/articles/dual-booting/

### LINKED ISSUE #462: Dual boot article has incorrect images
https://support.system76.com/articles/dual-booting/

The article begins with "If Pop!_OS is installed first", but then shows a drive in gparted with an NTFS Windows partition. I believe this image was in a previous iteration of the article in which Windows was installed on the drive first.

While you're in there, you may want to also fix that first paragraph. It says, "Once you're booted into Pop!_OS..." but does not mention that you have to either boot to a live disk or the recovery partition, as you can't resize your Pop partition while booted into that Pop install.

### LINKED ISSUE #532: Order of partitions for dual boot
Hello!! I have successfully installed PopOS on my laptop. Now I need to install Windows in dual boot. Although I understand the instructions to do it, I'm not totally clear from which "side" should I free up the space from the root partition. Specifically...should I reduce from left to right, so that the free space for Windows is adjacent to the Efi partition? Or should I reduce from right to left so that the free space for Windows is adjacent to the swap partition? Or is this not relevant and can be done either way?

I hope the image attached helps clarify my question.

Also, I'm not clear on how to choose which OS to boot after the installation is finished...do I just press the space bar to open systemd-boot and choose the system? No boot-fixing needed, as in other Linux distros? 

Thanks in advance for any help!

Captura de pantalla de 2021-01-22 21-21-33

jacobgkau:
- Closes #319 (clarifies that full-disk encryption is not supported when dual booting on a single drive.)
- Closes #462 (replaces incorrect images with valid screenshots.)
- Closes #532 (shows screenshots with valid examples for order of partitions; also explains how to select the default OS and switch between OS's, in a more clearly labeled section and with more detail.)
- Replaces mislabeled image added in #541 (also simplifies the WoeUSB process, as WoeUSB-ng switches to NTFS automatically when required.)
- Removes unused images that have been hanging around.
- Provides more step-by-step detail for the two covered scenarios.
- Replaces LCD photos with screenshots for easier viewing.


_articles/dual-booting.md
```diff
@@ -10,100 +10,350 @@ keywords:
...
 ---
 
-## DISCLAIMER
+## Installing on a Shared Drive (Starting with Pop!_OS)
+
+If you need to install both operating systems on a single drive (for example, if your computer only supports a single SSD), it is possible to install both OS's using a custom partition scheme. Pop!_OS's full-disk encryption is not supported with this setup.
+
+Start by [installing Pop!_OS](/articles/install-pop/), making sure to select `Don't Encrypt` during the installation process.
+
+(System76-imaged computers give you the option of whether or not to encrypt on first boot. If you chose to encrypt at first boot, or if you reinstalled and chose to encrypt, you will need to reinstall without encrypting before proceeding. If you want to dual boot and also use full-disk encryption, you will need to [use a dedicated drive](#installing-on-a-dedicated-drive) instead.)
```
ahoneybun, line 163: This is only true when Pop is installed and not Ubuntu so we may need to add that. The rest looks amazing.

ahoneybun (changes requested):

_articles/dual-booting.md
jacobgkau: This article is specifically about dual booting with Pop!_OS, but I've removed the ambiguity with https://github.com/system76/docs/pull/550/commits/7461589c7b45cc819d7ebc2316f53c62b8739413 - is the new wording good?

ahoneybun: Looks good I just want it as clear as possible.

ahoneybun (approved):

PR #551, opened 2021-02-23
Remove galp4 and darp6 PDF service manuals

jacobgkau:
The galp4 and darp6 manuals were redone from scratch in tech-docs, and the [service manuals article](https://support.system76.com/articles/service-manuals/) already points to tech-docs only for galp4 and darp6. This removes the PDFs to avoid users finding old, bad instructions. (Removing in response to a complaint about the keyboard removal process in the galp4 PDF manual.)

We also need to make sure this isn't linked anywhere on the website.


leviport (approved):

jacobgkau:
https://system76.com/guides/galp4
https://system76.com/guides/darp6

Links to the PDFs have been removed from the guides pages. The order page links to the guides page (and not directly to the service manual), so we should be covered here.

ahoneybun (approved):
LGTM thank you!

PR #555, opened 2021-02-24
Note lack of AMD GPU support

### LINKED ISSUE #490: "Install DaVinci Resolve" instructions with AMD graphics
### Issue
The current instructions for installing latest AMD drivers (20.40) required for DaVinci Resolve are not compatible with pop_os (20.04) or kernel version (5.8.0-7625-generic).

1. pop_os is unsupported
    ```bash
    $ sudo ./amdgpu-pro-install -y --opencl=legacy,pal
    Unsupported DEB-based OS: /etc/os-release ID 'pop'
    ```
   If you replace `pop` with `ubuntu` in /etc/os-release, then you get further...
1. kernel version unsupported by AMD driver kernel modules

- https://support.system76.com/articles/install-davinci-resolve/
- https://github.com/system76/docs/blob/master/_articles/install-davinci-resolve.md#install-prerequisite-packages

jacobgkau:
Closes #490.

It's currently not possible to get DaVinci Resolve running on Pop!_OS because Davinci Resolve requires the proprietary AMDGPU-PRO driver (not just the OpenCL portion that can be replaced by the open-source ROCm stack, but also the OpenGL portion); and AMDGPU-PRO does not work on kernel version 5.8.0 (it targets Ubuntu 20.04.1, which ships 5.4.0, and RHEL/CentOS; Ubuntu 20.04.2 is not supported yet: https://www.amd.com/en/support/kb/release-notes/rn-amdgpu-unified-linux-20-45)

This is fixing the documentation issue only. The issue of AMDGPU-PRO support on Pop!_OS is still being tracked here: https://github.com/pop-os/pop/issues/1135 

This replaces/re-purposes the section previously noting that Intel wasn't supported with a new section specifying that Resolve requires NVIDIA and doesn't work with Intel or AMD GPUs.

(It also updates the instructions to reflect the latest filenames and procedures for installing.)


leviport (approved):

curiousercreative:
Thanks for updating this. Such a bummer that neither AMD or Intel can run Resolve on Linux ATM :( I've migrated to Pop!_Os from macOS in large part to escape GPU vendor lock-in

Issue #406, opened 2020-06-21
Misleading Title of Installation Tutorial


chinmaya1011:
https://support.system76.com/articles/install-pop/ 

The name of this article gives the impression that the tutorial is only valid for System76-manufactured computers.

ahoneybun:
This has been merged so I'm closing this, thank you for the feedback!

PR #560, opened 2021-02-26
Update Install Pop name

ahoneybun:
This addresses this issue:

https://github.com/system76/docs/issues/406


storyaddict (approved):
Making this article more general purpose is a good idea.

PR #563, opened 2021-02-26
Add repair drive recovery

ahoneybun:


storyaddict (changes requested):
Good start. Let's add instructions for Live Disk/Recovery and link to those articles as well for cross-referencing.

Also don't need an Ubuntu AND Pop article for this, we can just put it under software and applications.

storyaddict (changes requested):
Let's just have one article "fixing drive" for both PopOS and Ubuntu.

ahoneybun:
The issue with that is that we have two sections for pop and Ubuntu due to the pop page.

_articles/fixing-drives.md
```diff
@@ -0,0 +1,36 @@
...
+
+---
+
+## Problem
+
+File Systems can be damaged from events like unexpected power loss or unsafe removal of drive(s). This is where you would run a File System repair tool on your system. There are a few ways to do this and we'll list them below:
+
+## Solution
+
+If you are not able to boot into the installed OS due to this issue we can boot into a [live disk](http://support.system76.com/articles/live-disk/), [mount then chroot](https://support.system76.com/articles/bootloader/) into the installed OS to go though these steps.
```
storyaddict, line 22: Grammar:

"mount the drive, then chroot into the installed OS. Once chrooted, follow these steps:"

storyaddict (changes requested):

_articles/fixing-drives.md
ahoneybun: Are you sure?

storyaddict: Yes. Mount [link] and chroot [link] should each have hyperlinks to those specific parts of the article. And the sentence is clearer in two parts.

Previously it was a run-on sentence.

storyaddict (approved):
Added the link I wanted to include.

PR #565, opened 2021-02-26
remove changing passphrase due to possible damage and to match the Ub…

ahoneybun:
…untu version, also removing GRUB steps


storyaddict (approved):
Removing Ubuntu section makes sense. Also adjusted wording to avoid accidental clicks of "install."

PR #517, opened 2020-12-27
Added ThinkPad Thunderbolt 3 Workstation Dock Gen 2

rubenbakker:
I'm using a ThinkPad Thunderbolt Workstation Dock Gen 2 (40ANY230CH) with a Galago Pro `galp5`.
Peripherals connected to the Dock:

* Dell 4K 24" monitor using DisplayPort. At work I tried the same dock with a Samsung 27" 4K monitor.
* Logitec USB receiver for a wireless MX Anywhere 2S mouse
* Wired USB Keyboard (Leopold FC660C)
* Ethernet
* Headphone using the 3.5mm Headphone Jack

No additional drivers were needed, everything just worked.

Only problem so far: The Galago Pro doesn't charge through the dock. [Link to related issue](https://github.com/system76/firmware-open/issues/144)

PR #567, opened 2021-02-27
Add Lemur Pro 10, replacement Lenovo dock

cipherboy:
I've recently purchased the Lemur Pro 10 and noticed it was missing from
the dock support page. I was also able to test with a newer revision of
an existing dock and it (still) works well.

Peripherals connected to the dock:
 - 2x Lenovo p24q-10 / 2560x1440 @ 60Hz over HDMI
 - 1Gbe ethernet
 - USB Hub (built into monitor) out to keyboard/mouse.
 - Headphones w/ mic (only works if plugged in prior to attaching USB-C)

Cables attached to computer:
 - USB C from dock

Charging from dock appears to worked well on this device, unlike in #517.

`Signed-off-by: Alexander Scheel <alexander.m.scheel@gmail.com>`


jacobgkau:
Thank you for the contribution. I'm making an additional commit to fix a couple of related things:
- Currently, galp and darp generations are listed in descending order. I'm moving the new lemp10 entry to the top so lemp is also in descending order.
- I'm moving darp above galp to at least keep the list in semi-alphabetical order. (I'm not sure where the current order came from.)
- I'm also adding galp5 and darp7, since they're current-generation machines not on the list.

jacobgkau (approved):

PR #568, opened 2021-03-03
update title and desc of live disk articles

ahoneybun:
This makes the article easier to find and understand what they are for.

### DIFF
--- a/_articles/live-disk.md
+++ b/_articles/live-disk.md
@@ -1,8 +1,8 @@
 ---
 layout: article
-title: Create and Use Bootable Media from Other OS's
+title: Live Disk Creation (Other Operating Systems)
 description: >
-  You can run Pop!_OS from a USB drive for hardware testing, recovery, and installation/re-installation.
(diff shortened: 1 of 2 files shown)


_articles/live-disk.md
```diff
@@ -1,8 +1,8 @@
 ---
 layout: article
-title: Create and Use Bootable Media from Other OS's
+title: Live Disk Creation (Other Operating Systems)
 description: >
-  You can run Pop!_OS from a USB drive for hardware testing, recovery, and installation/re-installation.
+  Create and Use Bootable on Other Operating Systems.
```
jacobgkau, line 5: Looks like you dropped a word here? ("Media" is still in the other description, but not this one. "Use bootable on" doesn't make sense.)

Also, you might want to get rid of the title capitalization in the new descriptions. None of the other article descriptions are capitalized this way.

storyaddict (approved):
Making the article titles more in line with how we refer to things in support tickets.

PR #570, opened 2021-03-04
Add oryp7 to Open Firmware table

jacobgkau:


ahoneybun (approved):
Live version of the site looks good! Thanks!

PR #572, opened 2021-03-15
Correct package name of typescript

syohex:
Both Ubuntu and debian use node-typescript

Reference
- https://packages.ubuntu.com/focal/node-typescript
- https://packages.debian.org/buster/node-typescript


jacobgkau (approved):
Thank you for the correction!

PR #574, opened 2021-03-17
Update Tensorman article and remove obsolete TensorFlow articles

jacobgkau:
Unless I'm misunderstanding something, the two TensorFlow articles that cover manual installation instead of using Tensorman are no longer relevant; we are no longer providing the `tensorflow-cuda-latest` package in Pop!_OS 20.10, and the message at the top of each article was meant to direct current users to the Tensorman article instead.

Additionally, there were a few issues in these articles:
- The Python API example in the "Use TensorFlow" article used `tf.Session()`, which is [no longer valid](https://stackoverflow.com/questions/55142951/tensorflow-2-0-attributeerror-module-tensorflow-has-no-attribute-session) in TF2.
- In the Tensorman article, there were two sections covering the command to install Tensorman.
- In the Tensorman article, there were a number of small grammatical and formatting errors (mostly missing commans and unclosed underline tags.)

To address the above issues, this PR updates the Tensorman article with some brief proofreading and a new section for the Python example. It sets the old article URLs to redirect to the new article. The description should make it easy to locate if someone's looking for TensorFlow-specific documentation.

I tried to also include the C API example from the old TensorFlow article, but it got to be long/involved enough that anyone who was able to follow along probably doesn't need it spelled out. If we think that particular section is very important, I can continue trying to add it.


storyaddict (approved):
Proofreading looks good. Thanks @jacobgkau! 

And I agree that we should pare down our TensorFlow/Tensorman and consolidate it here.

ahoneybun:
I believe this :

PR #575, opened 2021-03-18
add every product guide link and link to the service manuals

ahoneybun:


storyaddict (changes requested):
bown10 link doesn't work.
bown13 link doesn't work.
No content in daru4
gazp9 link doesn't work
kudp1 page has no content
oryx1 link doesn't work
serw8 no content
meer1 has no content
ratp2 has no content
sabc2 has not content
sabt3 has no content
wilp10 has no content

storyaddict (changes requested):
bonw12 link is dead.

what's the deal with gaze14-17a? Are we missing info on the guide page like for the other screen sizes, or is this all we have?

Same question for Oryp5-17

Same question for ser10-17

Same question for  leow9-core-xb

sabl6 is dead.

Service manuals link at bottom of article populates: 

https://github.com/system76/docs/blob/bd8d95390884d1932c5dd916162122e22aa56ff2/articles/service-manuals

storyaddict (approved):
Have checked links and they are good. Merging.

PR #576, opened 2021-03-18
fix Gen 4 to Gen 3 for the M.2 NVMe

ahoneybun:


storyaddict (approved):
Discussed change with Aaron. Just changing an incorrect spec. Merging.

PR #584, opened 2021-04-07
Suggest cron to work around open firmware limits

ChristopherRogers1991:


ChristopherRogers1991:
Looks like `@reboot` doesn't work as hoped - will update to run on a schedule.

jacobgkau:
If this only needs to run once at boot, a systemd timer to run on boot would make more sense than a cron job running every ten minutes, wouldn't it?

ChristopherRogers1991:
If there's a simple systemd method to run just at boot, I'd be all for that. The cron job as written was basically just a "good enough" approach, achieving the result I wanted with a simple one-liner setup.

In theory, cron's `@reboot` would have achieved the same effect as a systemd timer running at boot, but it didn't work. Possibly the command ran too early during startup, and it needs to run during a later phase? Systemd might give more control over that, but I would expect it to be a more involved setup process (though that could be my lack of experience with systemd talking).

ids1024:
>If there's a simple systemd method to run just at boot, I'd be all for that. The cron job as written was basically just a "good enough" approach, achieving the result I wanted with a simple one-liner setup.

It's not a one-liner, but you should be able to get something to run at boot with a systemd service file using `Type=oneshot`.

jacobgkau (approved):
I added a missing step that was required when testing this solution. Looks fine now.

PR #587, opened 2021-04-14
add support times

ahoneybun:
This adds our support hours on our support page (support.system76.com) per Alex on support's feedback.


Socialhappiness (dismissed):
Remove "at" before the word 'between'
NVM, Aaron was right.

Socialhappiness (approved):

PR #589, opened 2021-04-23
Update dead link for Pop!_OS's Roadmap article

Gadiguibou:
It seems the link has changed but was not updated in the process. I think that the link leading to "This Week in Pop!" should also be updated or removed since this doesn't seem to be active anymore.


ahoneybun (approved):
LGTM thank you for your work!

PR #592, opened 2021-05-05
Add jekyll-redirect-from plugin

burden:
This should enable the `redirect_from` blocks found in some of the pages front matter.


jacobgkau (approved):
Doesn't seem to affect my local installation (redirects worked before and after pulling this down.) We definitely want this to work in production, the gemfile looks valid and CI is passing.

PR #593, opened 2021-05-05
Update base url to https

jacobgkau:
This should make redirects (such as `/articles/use-tensorman` to `/articles/tensorman`) redirect to https, where they currently redirect to http even if accessed over https.


jacobgkau:
Not sure of a good way to test this locally, since a local build with `bundle exec jekyll serve` doesn't use https or the configured URL.

burden:
This would be the first step in ensuring that all visitors use `https://` which may or may not even be necessary if it's currently not being enforced. 

If we decide to merge this, be sure to enable the `Enforce HTTPS` feature at https://github.com/system76/docs/settings/pages first

jacobgkau:
I see two different issues:

- Accessing `support.system76.com` for the first time uses http by default. From what I understand, the `Enforce HTTPS` feature would address this.
- Accessing `https://support.system76.com/articles/use-tensorman` currently redirects to an http page. I consider this worse, since someone explicitly trying to use https gets redirected back to http. From what I understand, this PR on its own would address this.
    - (Any redirect via the Jekyll plugin, from http:// or https://, would redirect to https with this change; this should still work even if we don't set `Enforce HTTPS`, right?)

But enforcing https in general to address the first issue seems like it might not be a bad idea.

doomspork:
@jacobgkau / @burden do we still need this PR?

jacobgkau:
Just came across this PR again while going through old notifications. A website like https://redirectdetective.com/ shows that the redirects still go from https:// to http://, then back to https://. I think it would still make sense to merge this to get rid of that extra redirect-- since we're now enforcing HTTPS anyway, there should not be any downside to this.

btkostner (approved):

PR #596, opened 2021-05-18
Update bluetooth.md

twigglits:
Extra command for restarting bluetooth service.


jacobgkau:
This should probably be `sudo systemctl restart bluetooth` instead of the deprecated `service` command, and it might make sense to include it farther down alongside the `sudo systemctl status bluetooth` command already listed.

twigglits:
Yes, that makes much more sense. Thank you for correcting.

Issue #591, opened 2021-04-29
Clarify general applicability of graphics/power settings in docs


metasoarous:
The page on switching power/graphics profiles reads

> The following laptops have switchable graphics:
>
> Adder WS (addw1, addw2)
> Galago Pro (galp5 - NVIDIA models only)
> Gazelle (gaze14, gaze15)
> Oryx Pro (oryp4, oryp4-b, oryp5, oryp6)

Given that many people arriving at this documentation page may be using non-System76 hardware, it may be helpful to more clearly spell this out.

ahoneybun:
Thank you for the feedback! Are we looking to add System76 to the front of those products or something else? We can't confirm what models from other OEMs have working switchable graphics though.

metasoarous:
Hi @ahoneybun. Thanks for your response.

Somehow, I thought I had created a PR out of the commit referenced above, but never did. See #597 for my suggestion about how the general utility of the functionality could be made clearer.

Thanks again

PR #597, opened 2021-05-18
Clarify scope of switchable graphics support; solves #591

metasoarous:


jacobgkau (changes requested):
> Many laptops these days have _switchable graphics_, which allow users to switch between high performance and battery saving modes (as well as an intermediate "balanced" mode).

This wording makes it sound like High Performance, Balanced, and Battery Life are related to switchable graphics, which they're not. They are available on all machines. The modes for graphics switching are Integrated, NVIDIA, Hybrid, and Compute.

storyaddict (approved):
Adding Oryx7 and removing wording about "balanced" etc.

jacobgkau (approved):

PR #598, opened 2021-05-19
Update cuda.md

theofpa:
Fixes: system76/cuda#17


mmstick:
It may apply for a non-Docker system-local install of tensorflow, but not CUDA

ahoneybun:
@theofpa does this make sense? I think we can close this PR since this doesn't apply for our packaging.

storyaddict:
Closing per @ahoneybun's last message.

PR #603, opened 2021-05-20
Add brother article

Seibz:
I've added an article and screenshot on how to set up Brother printers to our add-a-printer article.


Socialhappiness (approved):
Looks good! The printer properties image may be broken. Double check image in article once it's live :)

PR #604, opened 2021-05-20
Log-in Loop Ubuntu - Maintenance

gangwerz:
- Added Thelio Mira to the Nvidia availability matrix
- Removed fix for Pop systems. There's another article for that
- Added bash language to code blocks for syntax highlighting


storyaddict (approved):
Changes look good.

PR #606, opened 2021-05-20
Split log file section from System76 Driver to new article

nathandyer:
Moved the log portion of the System76 driver page to a new article, and expanded on the information about log file creation to include doing so manually, and explaining what's included within the logs.


ahoneybun:
I was working on these changes in another PR.

storyaddict (approved):
Splitting these articles up makes sense, as well as having the Terminal version of log generation. Great work, Nathan.

PR #611, opened 2021-05-21
Update Ubuntu 20.04 Download Link to Fix 404 Error

nathandyer:
Updated the link to the Ubuntu 20.04 download, as the original link would load a download page, which would then redirect into a 404.

This new link goes to the download page, then presents the user with the download of the ISO file.


ahoneybun (approved):
LGTM! :tada:

PR #615, opened 2021-05-25
Update openvpn article

ahoneybun:
This adds steps for installing the ProtonVPN GUI client to this article since they released it publicly.


thomas-zimmerman (approved):
LGTM

PR #617, opened 2021-05-26
Add Fedora commands

Seibz:
Added commands for installing the system76-driver on Fedora, using the community-maintained repository [here](https://copr.fedorainfracloud.org/coprs/szydell/system76/)
Will add full instructions on installing all system76 packages in a different article.


storyaddict (approved):
Fedora additions look good, @Seibz, thanks!

PR #618, opened 2021-05-26
Update before-you-open-a-support-ticket.md

storyaddict:
Fixing a formatting issue that only showed up after the article went live.


ahoneybun (changes requested):
Looks good now.

ahoneybun (approved):

PR #620, opened 2021-05-27
reorder Support Tickets section

ahoneybun:
Here you go Nathaniel.


storyaddict (approved):
Putting "Before you open a ticket" article in the right place.

PR #623, opened 2021-05-28
add dkms enable command for Fedora

ahoneybun:
This allows the keyboard backlights to be controlled on Fedora on our systems.


storyaddict (approved):
These changes make sense. Adding the enable service and reboot steps will make sure the install completes and changes are applied.

PR #624, opened 2021-06-02
Added partition unmount to docs

Tim-Jackins:
I thought this instruction in the docs would help people use the WoeUSB-ng tool. Let me know what you think.


jacobgkau:
Before flashing, WoeUSB-ng should currently be ejecting the target automatically, or at least attempting to: https://github.com/WoeUSB/WoeUSB-ng/pull/24 

After flashing, ejecting manually should not be necessary if you're installing on the same device. If you're creating installation media for a different device, then ejecting the device instead of the partition shouldn't make a difference. What issue are you running into when you omit this step?

You also added additional subsections to the table of contents. The omission of subsections for the first few sections was intentional; the only reason I included subsections for the final section was to make common post-installation questions like installing drivers and repairing the bootloader more easily discoverable. I can't really argue against adding the rest, other than it making the table of contents ~50% longer and making it easier to jump to sections in the middle of the installation where it might not be as common to jump to.

Tim-Jackins:
Ok, yeah I tried to use the WoeUSB tool on ubuntu and popos and it didn't automatically eject the partition and just failed until I unmounted the partition manually. I may have been using it incorrectly but thought this was a possible troubleshooting step common enough to add here.

This was related to an issue with legacy booting that I got resolved with system76 tech support. The PR may not be needed anymore.

Thank you for your time!

PR #626, opened 2021-06-03
Updated alternate ways to enter recovery mode

IamDaedalus:
Updated alternate ways for users running only Pop!_OS to enter the Recovery partition if holding Space doesn't work for them


ahoneybun:
@jacobgkau I don't think those keys do anything on our systems.

IamDaedalus:
> @jacobgkau I don't think those keys do anything on our systems.

I am running only Pop!_OS on an HP and that usually gets me into recovery most of the times than holding space does

jacobgkau:
@IamDaedalus Thank you for the contribution! System76 documentation is primarily for customers and users of System76 hardware, although we of course want to help people run Pop!_OS on other hardware whenever possible.

@ahoneybun systemd-boot will actually start when most keys are pressed, which does include `F11` and `F12`. This would work on Open Firmware laptops because the firmware on those laptops only intercepts the `ESC` key. However, some of our desktops use `F11`/`F12` for the BIOS menu, boot menu, network boot, etc, so I wouldn't recommend documenting those keys as being systemd-boot keys. We use `Spacebar` because it's a key that the UEFI/BIOS splash screen typically won't care about.

storyaddict (approved):
Last commit addresses Jacob's and IamDaedalus' concerns.

PR #628, opened 2021-06-03
Add laptop power on article

ahoneybun:
This PR adds a new article similar to the Desktop article for powering on issues with troubleshooting steps to try before contacting support. Here is a preview of the page:

ksnip_20210603-111111


Socialhappiness (changes requested):
Change Title to 'Power on Failure- Laptop'
Since the page groups them in alphabetical, Power should be the first word of the Title

ahoneybun:
> Change Title to 'Power on Failure- Laptop'
> Since the page groups them in alphabetical, Power should be the first word of the Title

Fixed for this article and the desktop one in the latest commit.

Socialhappiness (approved):

PR #630, opened 2021-06-03
Freezing article

ahoneybun:
I think this article is pretty much done other then a once over.


storyaddict (approved):
Made my desired changes in the latest commit.

PR #632, opened 2021-06-04
Update software article

ahoneybun:
This adds the GNOME Shell Extension for System76 Power on Fedora.


storyaddict (approved):
See latest commit notes, but otherwise looks good!

PR #633, opened 2021-06-04
Add protonup

Seibz:
Created directions on using `protonup` to installer newer GloriousEggroll forks of proton.


ahoneybun:
@storyaddict do you think we should rename the article at this point?

ahoneybun (approved):
LGTM content wise

PR #635, opened 2021-06-04
Basic Networking Diagnosis article

linuxgnuru:


_articles/network.md
```diff
@@ -0,0 +1,313 @@
...
+The layers in the TCP/IP network model, in order, include:
+
+-   **Layer 5:** Application
+-   **Layer 4:** Transport
+-   **Layer 3:** Network/Internet
+-   **Layer 2:** Data Link
+-   **Layer 1:** Physical
+
+We will assume that you are familiar with this model, and will proceed by discussing ways to troubleshoot issues at stack Layers 1 through 4.
+Where to start troubleshooting is situation-dependent. For example, if you can SSH to a server, but the server can't connect to a MySQL database, the problem is unlikely to be the physical or data link layers
```
nathandyer, line 34: Missing a period at the end of this sentence.

_articles/network.md
```diff
@@ -0,0 +1,313 @@
...
+  - diagnosis
+image: http://support.system76.com/images/system76.png
+hidden: false
+section: software-applications
+
+---
+
+Finding out why your system is having network issues can be a daunting task.  We'll explore some useful commands 
+
+[]{#anchor}A quick review of the TCP/IP model
```
nathandyer, line 18: I'm not sure about the []{#anchor} format. Will this still make it an H1?

_articles/network.md
linuxgnuru: ahh, 2 lines should've been joined.

_articles/network.md
linuxgnuru: ugh; that was from pandoc

_articles/network.md
```diff
@@ -0,0 +1,313 @@
...
+
+**lo UNKNOWN 00:00:00:00:00:00 \<LOOPBACK,UP,LOWER\_UP\>**
+
+**eth0 UP 52:54:00:82:d6:6e \<BROADCAST,MULTICAST,UP,LOWER\_UP\>**
+
+It looks like **ip link set eth0 up** did the trick, and eth0 is back in business.
+
+These commands are great for troubleshooting obvious physical issues, but what about more insidious issues? Interfaces can negotiate at the incorrect speed, or collisions and physical layer problems can cause packet loss or corruption that results in costly retransmissions.
+
+How do we start troubleshooting those issues?
```
nathandyer, line 79: Should this be a heading? i.e., ## How do we start troubleshooting those issues?

linuxgnuru: changed

nathandyer:
Just a general note, I'm not sure about the formatting of commands, both in-line and on their own lines.

Rather than bolding/italicizing, should we do:

`inline command` 

```
code block
```
 
instead?

jacobgkau:
I've renamed the PR because "initial push" isn't useful when browsing GitHub notifications.

ahoneybun:
I would take a look at this page for formatting and also look at our other article to match better.

https://github.com/system76/docs/blob/master/CONTRIBUTE.md

_articles/network.md
```diff
@@ -0,0 +1,312 @@
...
+A quick review of the TCP/IP model
+---------------------------------------------
+
+First, let's take a moment to review the fundamentals of the TCP/IP network model.
+
+![TCP/IP OSI model](/images/network/tcpip.png)
+
+The layers in the TCP/IP network model, in order, include:
+
+-   **Layer 5:** Application
```
jacobgkau, line 27: Application is layer 7, not 5, and you skipped the Presentation (6) and Session (5) layers.

_articles/network.md
```diff
@@ -0,0 +1,312 @@
...
+  - networking
+  - command line
+  - diagnosis
+image: http://support.system76.com/images/system76.png
+hidden: false
+section: software-applications
+
+---
+
+Finding out why your system is having network issues can be a daunting task.  We'll explore some useful commands
```
jacobgkau, line 16: This sentence is not complete.

_articles/network.md
linuxgnuru: I was going off of the OSI model

jacobgkau: The OSI model has seven layers: https://en.wikipedia.org/wiki/OSI_model

ids1024: 5 layers is correct for TCP/IP, as it is typically discussed: https://en.wikipedia.org/wiki/Internet_protocol_suite#Comparison_of_TCP/IP_and_OSI_layering

jacobgkau: I see now that this section is labeled "in the TCP/IP network model," not OSI (although @linuxgnuru just said he was going off of OSI.) For TCP/IP, shouldn't the bottom two layers be collapsed into one as well? The TCP/IP model doesn't separate data link from physical.

https://bytesofgigabytes.com/networking/tcp-ip-model/ 
https://www.guru99.com/tcp-ip-model.html 
https://www.geeksforgeeks.org/tcp-ip-model/

ids1024: It seems there's some debate about what model to use for TCP/IP, but 5 layers is fairly common. Physical layer would be, say, 1000BASE-T over twisted pair copper wire, while the link layer deals in Ethernet frames that are independent of the exact physical medium.

The layers in a modern TCP/IP stack aren't all entirely independent.

jacobgkau: While I acknowledge some authors write five layers in TCP/IP-like models, the Wikipedia page you linked seems to describe four layers:

image

image

The page states that "Several authors have attempted to incorporate the OSI model's layers 1 and 2 into the TCP/IP model... This often results in a model with five layers," but also that "The IETF protocol development effort is not concerned with strict layering." and that "Some of the networking models are from textbooks, which are secondary sources that may conflict with the intent of RFC 1122 and other IETF primary sources."

This entire support article seems somewhat out-of-scope to me (I somehow doubt that someone who doesn't want to research these topics from more authoritative sources would take the time to read and understand the content here), but if it's going to be here, I would personally recommend using the OSI model since it's actually definitive. I suppose since the TCP/IP model isn't very well defined, I can't win an argument about the number of layers being incorrect, but it looks off to me.

The [image](https://github.com/system76/docs/tree/network/images/network) that @linuxgnuru included also lists four layers for the TCP/IP model, so if we're taking the stance that it's five layers, that graphic conflicts with the text of our article.

The next line, "We will assume that you are familiar with this model," also seems strange. What happens if the reader's not familiar with the model?

Socialhappiness:
'Finding out why your system is having network issues can be a daunting task. We'll explore some useful commands'
Don't plant negative words in documentation please. Please just describe the article in that section. For example;
'Useful commands for network troubleshooting'

storyaddict (approved):
Made formatting and a few wording changes for clarity.

_articles/network.md
jacobgkau: The image has been removed and the "assuming" line has been removed, so I'm marking this as resolved.

PR #636, opened 2021-06-05
Formatting consistency

storyaddict:
Standardizing formatting around document titles, bash excerpts, and authorship footers. Most articles have been updated. Doing a PR now to merge some other, higher priority changes after/on top of these.


_articles/codecs.md
```diff
@@ -70,6 +72,12 @@ After enabling codec support, it's recommended to restart your computer.
 
 The program <u>VLC</u> is a excellent media player with support for DVDs, CDs, and most formats of media files.  It can be installed with the the <u>Pop!_Shop</u> in Pop!_OS, in the <u>Software Center</u> in Ubuntu, or with this command:
 
-```
+```bash
 sudo apt install vlc
 ```
+---
+
+- Document Version: 1.5.0
```
jacobgkau, line 80: What's the point of sections like these?

- Document version doesn't really mean anything, it just gets incremented (theoretically) whenever the article's updated. How many times the article has been updated relative to other articles doesn't seem like super useful information.
- The last modified date is already displayed at the bottom of every article with the `jekyll-last-modified-at` plugin, which updates automatically instead of needing to be manually updated in every PR to the article.
- The author and co-author/"editor" information is already available in the GitHub commit history, once again automatically as opposed to being a list that needs to be manually updated.

storyaddict: - Document Version shows how much or how little, at a glance, has been changed. It doesn't matter relative to other help articles (unless we're doing a full review of all of them).

- Last modified is not specific in that it could have been adding a missing comma. Version numbering and changing the date to match version updates is more significant information. 

- You shouldn't have to go to Github to see who has worked on an article. Authorship and attribution should be included so people can be publicly be acknowledged for their hard work without having to go to a separate page. This is the equivalent to having authorship printed on the front page of a book, instead of being buried in the credits at the end of a movie. 

- Emma and the careteam have already had this discussion and approved adding this footer.

- There were other formatting corrections made in this pull request too.

Socialhappiness: Here's my stance- I think it's important to take Jacob's feedback into account since he raised the same points I did.
- The fact that Jacob also brought up the same arguments makes me question the change.
- I still do not understand what problem is being solved by this- you guys made the argument that it made the article seem more human. But was it really ever a problem? I honestly don't think  the job of documentation is to show that we're more human. We do that on phones and in tickets.
- Document version creates more questions than answers
- It still just looks odd to me in the article. Last modified date alone looks fine and makes sense.
- The right people know that you wrote it and will check github. Your hard work doesn't go unnoticed just because your name isn't at the bottom of it. But I can see how it would feel that way.

Socialhappiness: I've discussed this in depth with Bjorn and he would like the footer to stay with the last modified date only. Please update.

storyaddict: I'll be commenting all the footers additions out.

PR #641, opened 2021-06-09
Update to Recovery article

thomas-zimmerman:
Change some of the table layout to only show different commands to run. Also simplified the bash command to mount system directories.


storyaddict (approved):
Made changes to both articles. Fixed a typo in hardware diagnosis. Expanded and clarified formatting in Recovery article.

PR #646, opened 2021-06-10
Fix formatting

ahoneybun:
This PR fixes the off center list for some of the sections like in the screenshot below:

ksnip_20210610-072555


jacobgkau (approved):
If the left-justified within column headings are expected, then this fixes the issue of some headings being left-justified to page instead:
Screenshot from 2021-06-10 07-36-57
Screenshot from 2021-06-10 07-37-07

Please feel free to request our review on major changes to avoid glitches like this.

PR #650, opened 2021-06-11
Top buttons

ahoneybun:
This PR adds buttons under each section to move you back to the top of the page where the section buttons are.


storyaddict (approved):
Good quality of life addition.

PR #658, opened 2021-06-13
NuxtJS Rewrite

btkostner:
This is a rewrite of the docs backend to [NuxtJS](https://nuxtjs.org/) and [@nuxtjs/content](https://content.nuxtjs.org/), the same backend the rest of the web projects are. ~Right now we are at feature parity. No new features were added, but this will make it much much easier to add new features like search and what not.~

Most of this PR is git junk for renaming all of the markdown files, and the static content. The actual code to this is a pretty small change. But here are some things you might notice
- Uses the same omnibar, header, and footer as the pop.system76.com page. No more Pinterest link. Yes, now it has an omnibar.
- Button styles are more in line to what system76.com has now
- Typography has changed a bit. I didn't port over the old typography, and I didn't use the new style guide headers because they don't make sense in an article format (header 4 is larger than 2 and 3.) It still uses our font though. And this should be easy to change later once UI/UX has free time. 
- There is now a search bar on the homepage. There are some weirdnesses about it, but I wrote it in an hour and we can improve on it later.
- I moved the "Edit on GItHub" button from the header to a footer on articles. Plus useful contributor info.

README has been updated with new file paths. The way we place icons like the Ubuntu and Pop!_OS logo has changed, so you might want to check that out.


btkostner:
@ahoneybun Aaron... I added basic search... You're welcome.

ahoneybun (dismissed):
I love the new look of the support site and the search box! Though there is a lot of spacing on the main page like the screenshot below:

ksnip_20210613-151034

If we can change that a bit I think we can merge and deploy it.

btkostner:
What spacing is the issue? The articles are lined up in a grid

ahoneybun:
I think between the subsections there is too much space but maybe that's just me. Technically everything looks good to me though.

btkostner:
@ahoneybun updated to be a bit smaller

image

ahoneybun:
Now a little less space between the sections and it will be perfect!

btkostner:
That's the last style change I'm going to make. If it functions, let's merge. It's easy enough for you to change and mess with. All of the style classes are from [tailwind](https://tailwindcss.com/docs/margin) and can be viewed in the page file /pages/index.vue:170. Keep in mind, these pages will probably change once UI / UX gets time.

ahoneybun (approved):
LGTM and I love the changes! :rocket:

PR #661, opened 2021-06-14
fix absolute urls in content files

btkostner:
Please stop linking to things with `https://github.com/system76/docs/blob/gh-pages` and `https://github.com/system76/docs/raw/gh-pages`. All of those files are served with the website, you can just do `/pdfs` or `/files/firmware` instead.


btkostner:
This doesn't fix the linking problem. This just addresses some of the urls being the full urls. Breaking this down into smaller PRs.

burden (approved):
:crown:

ahoneybun:
DO NOT MERGE!

ahoneybun:
> DO NOT MERGE!

We can merge this since we are using PR 663 for the main issue.

PR #662, opened 2021-06-14
Added updating PATH command to the article so protonup is available

gangwerz:
The Steam article had a section on using `protonup`, but it is installed to `~/.local/bin` which is not in the `$PATH` by default. This adds a line to do so for Bash users. Anyone else I would expect to be able to replicate the setting for their shell of choice.


gangwerz:
A better, alternate approach was mentioned. Abandoning this branch to pursue the alternative

PR #663, opened 2021-06-14
fix absolute path markdown links for static files

btkostner:
This should fix the absolute links not working for static file assets.


ahoneybun (approved):
This fixes the main issue without updating all of the articles. LGTM :taco:

PR #664, opened 2021-06-14
Added updating PATH command to the article so protonup is available

gangwerz:
On Pop/Ubuntu, `$HOME/.local/bin` is not added to `$PATH` unless the directory exists. Sourcing `$HOME/.profile` will add the folder to `$PATH` for the current session.


ahoneybun (approved):
LGTM :taco:

PR #665, opened 2021-06-14
Update and correct tables

ahoneybun:
This PR updates some article tables so that they work in the new backend.


storyaddict (approved):
fixing the tables

PR #666, opened 2021-06-15
move old hidden articles

ahoneybun:
This PR will move hidden articles to the archived folder since they are showing up in the search box now.


storyaddict (approved):
moving files

PR #671, opened 2021-06-15
Update pop recovery

ahoneybun:
This goes over using the Update option for the Pop Recovery in Settings -> OS Upgrade.


gangwerz (approved):
Looks good!

PR #673, opened 2021-06-16
fixing table spacing on shipping article, fixing lack of formatting o…

storyaddict:
…n format drive article, moved windows article to proper section.


ahoneybun:
Do not edit the html files only the md files.

PR #682, opened 2021-06-18
Fix article formatting

ahoneybun:
This PR fixes a few article either from not showing correctly or tables not working correctly.


storyaddict (approved):
Aaron and I reviewed this together over screen-share. Changes look good.

PR #683, opened 2021-06-18
add images for Install Ubuntu articles

ahoneybun:
This PR corrects missing images for the Install Ubuntu article.


storyaddict (approved):
In the process of moving the contents of one PR to the other, we didn't realize that the images were not moved to master. This corrects that.

PR #687, opened 2021-06-18
move article to the right place

ahoneybun:


storyaddict (approved):
file now in right place

PR #688, opened 2021-06-18
Linux gaming article

ahoneybun:
This PR adds a replacement article for the Steam article and also adds sections for GameHub and Lutris. It goes over installing those tools and using them to a degree which can be updated with more content later on.


content/linux-gaming.md
```diff
@@ -0,0 +1,185 @@
...
+### Install Lutris From Command Line
+
+Open the <u>Terminal</u> application by searching for <u>Terminal</u> after pressing the Super Key <kbd><font-awesome-icon :icon="['fab', 'ubuntu']"></font-awesome-icon></kbd>/<kbd><font-awesome-icon :icon="['fab', 'pop-os']"></font-awesome-icon></kbd>/<kbd>SUPER</kbd>.
+
+![Activities Overview](/images/linux-gaming/search-terminal.png)
+
+Once the <u>Terminal</u> application is opened you can use the Command Line tool `apt` to search for it like so:
+
+```bash
+apt search steam
```
jackpot51, line 144: Should this be lutris instead?

content/linux-gaming.md
```diff
@@ -0,0 +1,185 @@
...
+sudo apt install com.github.tkashkin.gamehub
+```
+
+**Be very careful when using sudo with ANY Command. It can make system wide changes so be sure to read everything before entering 'Y'.**
+
+### Install GameHub From the Pop!_Shop
+
+Open the <u>Pop!_Shop</u> application then either search for GameHub or by clicking the <u>GameHub</u> icon on the Pop!_Shop home page. Now click the **Install** button.
+
+![Pop!_Shop Steam](/images/linux-gaming/pop-shop_gamehub.png)
```
jackpot51, line 129: Should the alt text be GameHub?

ahoneybun: Fixed.

content/linux-gaming.md
ahoneybun: Fixed.

ids1024:
I suppose it would help to have some explanation of what `Lutris` and `GameHub` are and why one would want them? Or Steam, but that should be familiar to people who are already familiar with gaming on Windows.

storyaddict (changes requested):
We need to flesh out the "installing .exe" section. 

Also we need to explain more about Lutris and/or the differences between Lutris and Gamehub.

Should also highlight open-source games.

ahoneybun:
Yea I was working on that before I left the office.

ahoneybun:
@jackpot51 do you know how to fix the conflict? Their commands won't work since we can't push to master without a PR and the steam article is removed with this PR.

storyaddict (approved):
Updated.

PR #689, opened 2021-06-19
setup dynamic social images for articles

btkostner:
This will dynamically build social media images for articles based on the title. The old image files would look weird because they were not the correct resolution for twitter and what not. Now we have some cool photos that looks like this (screenshot cropping is a little off):

Screenshot from 2021-06-19 15 10 45

To test, you can run `npm run build` and check the `dist/_nuxt/images` directory.

Also (I know this is bad git work buuuuut):
- I removed `layout: ` because it does nothing
- Added in the `facebookImage` and `twitterImage` to each article file which might have caused a bit of whitespace trimming from my editor
- Removed the uppercase variants of `redirect_from` because everything gets down cased when redirecting


ahoneybun (approved):
LGTM everything from tables, to images and file downloads are working.

PR #691, opened 2021-06-20
Delete archive folder

btkostner:
Everything in this repo is in git. There is no reason to keep unused files in an archive folder when we can just go back in history to find them again. Plus, this will make the building action faster because we shallow clone it.


ahoneybun:
These articles are there as we can't hide them on the main site with the change so this folder will stay where it is.

btkostner:
What? They are not displayed on the site because they are not in the content folder... They are _never_ used in the build step. They are just dead git weight...

ahoneybun:
It is searchable on the search box on the site.

btkostner:
Ummmm, nope. The pages were never even built. I'll open another PR to fix that.

ahoneybun:
Yes support peeps saw it happen before we moved some articles to the archived folder.

btkostner:
They must have been using a cached version of the search database, because it doesn't build the pages, nor is it included in the search database.

btkostner:
@ahoneybun What is the intended effect here? Do you want it to have its own page? Do you want it in search results? Do you want it on the homepage?

ahoneybun:
No to all of those we're keeping them for historical reasons.

btkostner:
.... then why are they still here? We can delete this folder and still find them in git history...

ahoneybun:
The content is managed by support so that's a good enough reason to leave them there.

PR #692, opened 2021-06-20
Create a table of contents for large articles

btkostner:
This PR adds a feature that creates a table of contents if `tableOfContents: true` is set on an article.

Screenshot from 2021-06-19 21 39 39
Screenshot from 2021-06-19 21 39 58

Things to note, the headers extracted are a little weird because it's for h2 and h3 tags, so h1 tags are excluded. We should probably go over the articles and clean up the headers to correct levels.


ahoneybun:
@storyaddict I think this would be a nice addition, what do you think?

storyaddict (approved):
Love it!

PR #694, opened 2021-06-21
Update can't log in

storyaddict:
adding section about decryption issues, and how to fix using the Recovery partition.


content/login-loop-pop.md
```diff
@@ -132,6 +154,44 @@ sudo update-initramfs -c -k all
...
+sudo dpkg --configure -a
+sudo apt install -f
+sudo apt full-upgrade
+sudo apt autoremove --purge
+```
+
+Then:
+
+```bash
+sudo apt install --reinstall plymouth gdm3 gnome-shell pop-desktop
```
ahoneybun, line 181: I would also reinstall the kernel and rebuild the initramfs

apt install --reinstall grub-efi-amd64 linux-generic linux-headers-generic
update-initramfs -c -k all

ahoneybun (changes requested):
Needs a bit more but almost there.

content/login-loop-pop.md
storyaddict: Added.

PR #695, opened 2021-06-21
fix headers not being in markdown format

btkostner:
Doing `<br>` before every header was breaking the headers (even in GitHub markdown view.)

As a side note, please don't do formatting like this in markdown. If you want to have a horizontal line above every header, just ask web. We can format it that way and it will work consistently for every article (not just the ones you edit.) Plus it keeps the markdown nice and clean and easy to read.


ahoneybun (approved):
Looks good to me.

Issue #698, opened 2021-06-23
In-line code blocks are not rendering


gangwerz:
The single back-tick sections in the support articles are not being rendered properly. They now show in bold, with the back-ticks showing as part of the text

> > I am not seeing anything wrong with the article. However, the new Markdown to HTML compiler/transpiler has not been rendering in-line code (single back ticks), so any of those sections may need to be rewritten.
> 
> Should I just make it bold or do we have an option for an equivalent feature?

I think we should share with the team, cause it may require updating most articles, or another update to the markdown processor
Example on the live site: https://support.system76.com/articles/terminal-basics/

_Originally posted by @gangwerz in https://github.com/system76/docs/issues/678#issuecomment-867154597_

gangwerz:
@storyaddict @ahoneybun Tagging you both to bring eyes to this

btkostner:
This has been fixed in our styles. Thanks for the report @gangwerz!

PR #699, opened 2021-06-24
Setup basic code highlighting styles

btkostner:
@storyaddict ask and you shall receive (sometimes.)

- This updates inline code styles to match the code block styles.
- Updates code block highlighting to `solarized-dark` but with Pop!_OS colors. I only did some, but you can edit more of the colors in `/modules/content-highlighter.js`. Values are in `/assets/styles/code-highlighting.css` all stolen from Pop!_OS terminal colors. Shouldn't need to change that file.
- Tweaks the keyboard `kbd` styles a bit to make them more different from code blocks

I used http://localhost:3000/articles/use-tensorflow to check styles as it has the most different types of code languages. This probably won't be the last style change for code blocks, but it makes them better than before.

Also, any code block without a language hard set defaults to `text` which will not do any highlighting.


btkostner:
Fixes #698

storyaddict (approved):
Thank you very much @btkostner ! :D

PR #700, opened 2021-06-24
setup nuxt-picture for guide articles

btkostner:
This replaces images in markdown articles with `nuxt-picture` which during the build process will resize all pictures, optimize, and create a webp version.

TODO:
- [x] some image paths are missing


btkostner:
@system76/support This fixes all the broken images seen here https://github.com/system76/docs/runs/2901550731?check_suite_focus=true (builds step)

btkostner:
The two images deleted in the PR are variations of the ones we are actually using in the article, but are not linked from anywhere.

storyaddict (approved):
picture changes look good, nuxt file update accompanies it.

PR #703, opened 2021-06-24
added iCloud section.

storyaddict:
icloud content was not added correctly before the last merge. Adding that now.


ahoneybun (approved):
LGTM now! :tada:

PR #704, opened 2021-06-24
Fail builds if there is an error during build

btkostner:
Nuxt by default will just log errors. We seem to be missing them, so this will just fail CI or Deploy if there is an error logged.


storyaddict (approved):
These changes make sense.

PR #716, opened 2021-06-28
Remove initial power on disassembly instructions

Socialhappiness:
Article recommended disassembling machine as first step in troubleshooting. We don't want to encourage that when we can diagnose hardware failures with the steps outlined in this article on a per component basis.


storyaddict (changes requested):
Counteroffer - let's move the "won't power on" section to the bottom, put a note at the top and/or rely on the table of contents to say something like: "if your system won't power on, go here..."

We also have these:

https://support.system76.com/articles/power-on-failure-desktop 

https://support.system76.com/articles/power-on-failure-laptop

Socialhappiness:
We can add this at the bottom:
If the system fails to power on, please use the following articles to troubleshoot;
https://support.system76.com/articles/power-on-failure-desktop
https://support.system76.com/articles/power-on-failure-laptop

Socialhappiness (commented):
Adding if the system boots is unncessary. This article assumes the system will power on and there's a message to link to those articles later if that's not the case.
I don't like the component removal section. Too many new customers out of box go straight for the attackwith that troubleshooting and this has lots of room for failure. Especially wihout linking to instructions on the product parts repairs.

storyaddict (approved):
Made changes discussed with Emma.

PR #718, opened 2021-06-28
Revise main page

ahoneybun:
This PR address some feedback from Emma:

1. The bold formatting needs to remain consistent and within brand style guidelines
2. The clear cache and cookies article needs moved to troubleshoot section. It doesn't make sense to be right there.
3. The Learn section needs reworked- should it really be at the top of the page? Are people coming here to troubleshoot or to learn? It seems like there's excessive scrolling to get to the basics of troubleshooting and makes it harder to get to for customers if they miss that button up top
4. Each section shouldn't start with Learn about your operating system, learn about your computer. It can have a main heading as Learn, but not all the subheadings. Simplify it all into one section or suggest something different than 'Learn about' at the beginning of everything


btkostner:
I wouldn't worry about the bold formatting and brand guidelines as much. We're planning to completely change the typography on this site soon. It will be the first site to use the new typography that Maria is working on, hopefully done by next week. Other than that, the code looks good.

ahoneybun:
Then I guess the reorder of sections is the biggest change then @btkostner .

btkostner:
Yep Yep, and it looks like you know how to do all that, so :confetti_ball:

btkostner:
Also worth mentioning, at some point I'm probably going to do a little redesign of the header area to make the search bar more prominent, and the actions (like support tickets) a bit bolder.

btkostner (approved):
Code looks good to me

Socialhappiness (approved):

PR #719, opened 2021-06-28
Fix markdown linting

btkostner:
This is a huge PR to make the `Lint (Markdown)` check green. It mostly consists of:

- Removing the `h1` header at the top of the file
- Changing headers up or down a level so they all go in order
- Updating some indentation for lists to display right
- Fixing new line whitespace (auto done with `--fix`)
- Adding alt text to images
- Adding some missing / broken icons in the docs
- Changing some HTML (like `<strong>`) to the markdown version

Code changes consist of
- Adding back the article title to `/article/` pages after removing them from the markdown
- Changing a markdown lint rule to allow same headings in different nestings
- Removing a no longer needed Vue plugin (after renaming `<packagename>` to `[packagename]`)
- Adding icons to font awesome needed in some articles

Very little content should have actually changed, but all of the articles should look better now, especially if they have a table of contents showing.


ahoneybun (approved):
LGTM and I really enjoy the new changes at the top of the articles. The back button text saying Support Articles makes much more sense!

ahoneybun:
@btkostner I think the changes I just made are fine, is that correct?

btkostner:
Yep, changes look good!

PR #720, opened 2021-06-28
correcting link formatting at the end of hardware diagnosis article

storyaddict:


thomas-zimmerman (approved):
Should fix the images

PR #721, opened 2021-06-29
Updated to match new workflow in Pop!_OS 21.04

nathandyer:
Added the new username and password setup screens, as well as the change to the encryption password workflow, due to the changes in the 21.04 installer.

Also includes all-new screenshots.


ahoneybun (approved):
LGTM though let's merge at 4pm MT?

PR #723, opened 2021-06-30
update content and add GUI screenshot

ahoneybun:
This PR adds a screenshot of the OS Upgrade & Recovery section in Settings and reminds users to update the Recovery partition first before upgrading the OS. It also updates the description of the article to remove System76 since Pop can be installed on any x86_64 PC.


Socialhappiness (approved):
thanks for updating, looks good!

PR #728, opened 2021-07-07
update shortcuts and add workspace shortcut

ahoneybun:


storyaddict (approved):
Changing Activities to Workspaces makes sense.

PR #730, opened 2021-07-11
Minor grammar changes

tangowithfoxtrot:
Changed "a" to "an" when referring to ovpn files

### DIFF
--- a/content/use-vpn-software.md
+++ b/content/use-vpn-software.md
@@ -98,7 +98,7 @@ Access OpenVPN setting in the <u>Settings</u> application (GNOME Control Center)
 
 ### Getting an OpenVPN (.ovpn) file
 
-OpenVPN uses `.ovpn` files for setting up the VPN on your system and includes the servers to contact. Most VPN providers will provide  `.ovpn` files such as the following:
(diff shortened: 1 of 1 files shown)


jacobgkau:
I believe the reason "a" was used is because the "dot" in `.ovpn` is pronounced. You would say "a dot," not "an dot." That said, the ambiguity could be avoided by removing the leading `.` where possible and rewording the remaining instances.

tangowithfoxtrot:
Fair enough. I always skip the "dot" in my head, but I can totally see it either way. Would changing the `Getting a .ovpn file` line to something like `Getting an OpenVPN (.ovpn) file` be agreeable, or would you prefer it be something like `Getting an ovpn file`?

jacobgkau:
> Would changing the `Getting a .ovpn file` line to something like `Getting an OpenVPN (.ovpn) file` be agreeable

That suggestion looks great to me for the headers, as long as we either roll back the changes to the body text or reword it similarly in the body.

tangowithfoxtrot:
I've submitted new commits with revised grammar edits. Let me know if any further revision is necessary.

jacobgkau (approved):

PR #736, opened 2021-07-21
Update DisplayLink driver information

### LINKED ISSUE #735: Docking station article seems to imply that DisplayLink software is always needed for docking stations
Perhaps I'm reading it wrong, but the docking station article seems to imply that DisplayLink drivers are needed for docking stations. This isn't true. Docking stations that use normal standards won't need any additional drivers at all, whatsoever. DisplayLink drivers are only needed for docking stations that are specifically built with that technology.

(As a personal opinion, I recommend that everyone avoid DisplayLink altogether, it just adds challenges, and there are dozens of docking stations available that don't use that technology).

I'll be creating a video soon on my channel that will go over docking stations in more detail, so when I do that, it may be a good idea to update this article. But for now, you may want to re-word the DisplayLink section, installing that driver is probably never a good idea unless the device has a DisplayLink designation on it.

Cheers!

jacobgkau:
Closes #735, updates the version numbers/filenames in the example commands, and makes a couple of small wording improvements for clarity. Also moves the three disclaimers above the DisplayLink section since they apply more to the device lists higher up on the page.

### DIFF
--- a/content/docking-station.md
+++ b/content/docking-station.md
@@ -69,35 +69,35 @@ Community members have reported that the following docks work with our products:
   - Ethernet and DVI ports not tested.
 - [Plugable UD-ULTCDL Dock](https://plugable.com/products/ud-ultcdl/) [[community-tested](https://github.com/system76/docs/pull/518) on an NVIDIA system]
 
-## For Intel systems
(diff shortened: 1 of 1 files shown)


ahoneybun (approved):
LGTM and these are some great changes! Thanks!

PR #737, opened 2021-07-22
Added GRUB installation documentation

FlorianBreitenlacher:
Added GRUB installation documentation because some people want to use GRUB as their boot loader and it took me a few days to figure out how to install GRUB so why not share it and help people save time and nerves.


jacobgkau:
Thank you for the contribution; however, I'm not sure we want to provide instruction for installing a non-default bootloader when our engineering team will not fix issues with GRUB, our QA team will not catch bugs with GRUB, and our support team will not be prepared to support GRUB on Pop!_OS.

> because some people want to use GRUB as their boot loader

Why did you need/want to switch to GRUB from systemd-boot?

If this were to move forward, the operating system name is `Pop!_OS`, not `Pop! OS !`, and we certainly don't want to be linking outside YouTube videos in this documentation.

ahoneybun:
I agree with @jacobgkau I don't like having this in official documentation since it's not our default bootloader. We *could* have this as a community article with a big disclaimer that this is not tested or supported by our teams.

FlorianBreitenlacher:
@jacobgkau 
I fully understand the points you made and agree with them.
The reason why I wanted to use GRUB is because of the boot manager with the GRUB menu, which allows me to choose my operating system (for example Windows 10 and Pop!_OS). As far as I am aware, the Pop!_OS bootloader does not support such a feature. Maybe I am wrong and overlooked it in the documentation.

@ahoneybun
I also agree with you that the official documentation might be the wrong place for this.
If there should be a need for this documentation, I would like to post it on reddit or as a community article, with disclaimer included.

jacobgkau:
> The reason why I wanted to use GRUB is because of the boot manager with the GRUB menu, which allows me to choose my operating system (for example Windows 10 and Pop!_OS). As far as I am aware, the Pop!_OS bootloader does not support such a feature. Maybe I am wrong and overlooked it in the documentation.

There are instructions to switch OS's using systemd-boot, including making the menu appear on every boot without having to hold down any keys, under the `Using systemd-boot` portion of the `Switching between Pop!_OS and Windows` section: https://support.system76.com/articles/dual-booting#switching-between-pop_os-and-windows

FlorianBreitenlacher:
Yes, there was no Windows entry. Might be because I have Windows and Pop!_OS installed on different drives. I fixed it be copying the Bootloader. Now its showing up and let me boot into Windows or Pop!_OS.

FlorianBreitenlacher:
I think, I can close the pull-request now. Thanks for helping me out.

jacobgkau:
> Might be because I have Windows and Pop!_OS installed on different drives. I fixed it be copying the Bootloader.

Glad that you figured it out. If you'd like to try adding information on how to copy the bootloader, that might be a better solution for the documentation.

PR #741, opened 2021-08-06
Update guides manuals

ahoneybun:
This PR does the following:

- Updates the guides page with newer models
- Fixes the guides list with missing models
- Kills the Service manual article as it replaces it
- Creates one article to update instead of two


gangwerz (approved):
Looks good!

content/service-manuals.md
```diff
@@ -1,45 +0,0 @@
----
-title: Service Manuals for System76 machines
```
leviport, line 2: @ahoneybun did you mean to delete this?

ahoneybun: Yep compressing into one article so it's easier to keep up to date. I need to update pdf links in the next PR.

PR #743, opened 2021-08-09
update links

ahoneybun:
This PR does the following:

- Updates links to service manuals where they are available
- Tested each link and they all work


gangwerz (approved):
Looks good!

PR #744, opened 2021-08-10
docs: add apple music web client

pbnj:
Apple Music is accessible via web client at https://music.apple.com.


storyaddict (approved):
Thanks @pbnj for highlighting this.

PR #746, opened 2021-08-10
add github link for Open Firmware/Open EC systems

ahoneybun:
This PR does the following:
- Adds a link to the Windows drivers for our Open Firmware/Open EC systems


ahoneybun:
Do we want to add the other Thelio models?

jacobgkau (approved):

PR #749, opened 2021-08-12
Test i-tec USB 3.0/USB-C Dual HDMI Docking Station

pspacek:


pspacek:
I'm not sure whether I should point out that DisplayLink process consumes significant amount of CPU time when e.g. playing video on an external monitor. Maybe it is "expected" and behaves that way with all devices?

jacobgkau:
> I'm not sure whether I should point out that DisplayLink process consumes significant amount of CPU time when e.g. playing video on an external monitor. Maybe it is "expected" and behaves that way with all devices?

DisplayLink is software rendering, so I would probably call that expected behavior for DisplayLink. If we wanted to warn about that, it should probably be somewhere that applies to all of the DisplayLink docks.

jacobgkau (approved):
Thank you for the compatibility report!

PR #753, opened 2021-08-17
Update install-davinci-resolve.md

storyaddict:
Hiding Da Vinci Resolve article. Newer versions are broken.


ahoneybun (approved):
LGTM this is the best we can do for now.

PR #754, opened 2021-08-17
Software troubleshooting

storyaddict:
This PR moves software troubleshooting triage out of the "Before You Open a Support Ticket" article, and into its own. I've added a link to the new article on the "Before..."

### DIFF
--- a/content/software-triage.md
+++ b/content/software-triage.md
@@ -17,8 +17,6 @@ section: getting-help
 tableOfContents: true
 ---
 
-# Software Troubleshooting Triage
-
 ##  Gather Tools
 
 - [Backups]((/articles/backup-files))
@@ -27,9 +25,9 @@ tableOfContents: true
 
 ## Suspend, Reboot, Shut Down
 
-Do the symptoms go away if the system is suspended, rebooted or shut down and then started up again? Sometimes repeated suspends or reboots are all that are needed to clear an issue out. It's also a good idea to test reboot and shutdown separately as they handle memory and power-management differently. 
+Do the symptoms go away if the system is suspended, rebooted or shut down and then started up again? Sometimes repeated suspends or reboots are all that are needed to clear an issue out. It's also a good idea to test reboot and shutdown separately as they handle memory and power-management differently.
 
-## Symptoms 
+## Symptoms
 
 - System or Applications Crashing or [Freezing](/articles/freezing-behavior)
 - Black Screen or No Login Screen
@@ -103,7 +101,6 @@ You can open the terminal by clicking Activities in the upper-left corner and se
 When you run the first command with "sudo" it will ask you to enter your password.
 Typing your password will not show anything in the Terminal window. That is normal and for security.
 Just enter your password as you normally would, and then hit <kbd>ENTER</kbd>. You'll also need to hit <kbd>ENTER</kbd> after each command.
-
 
 ```bash
 sudo apt update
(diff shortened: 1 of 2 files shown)


content/software-triage.md
```diff
@@ -0,0 +1,316 @@
...
+facebookImage: /_social/article
+twitterImage: /_social/article
+
+hidden: false
+section: getting-help
+tableOfContents: true
+---
+
+## Gather Tools
+
```
gangwerz, line 21: Fix syntax typo's in the links

gangwerz (changes requested):
Typo's in the "Gather Tools" links

content/software-triage.md
storyaddict: Fixed.

content/software-triage.md
```diff
@@ -0,0 +1,311 @@
...
+### 4. Update with Packages Held
+
+To perform an update while keeping some packages on their current version, use the following Terminal command:
+
+```bash
+sudo apt-mark hold [packagename]
+```
+
+If the package is not central to OS functions, it can be held at the current version and the other packages updated around it. If the problem doesn't present itself when a package has been held, it was likely being updated to a newer version that was causing instability.
```
jacobgkau, line 137: I would caution against including holding a package this high up in the troubleshooting process.
- For one thing, you haven't explained how to downgrade a package yet, so if someone's having a problem and they follow this step, they would be holding the version that was causing the problem.
  - You also haven't explained how to identify which package to downgrade/hold.
- Also, holding packages is almost never a permanent solution. It should usually only be a temporary workaround. (If it's working around a bug, the bug needs to be reported/fixed; if it's working around a workflow change, the user needs to adapt their workflow to the new version before their old version stops working.)

storyaddict: I agree with this. The use of holding packages could be made clearer. I'll work your feedback into a revision.

content/software-triage.md
```diff
@@ -0,0 +1,311 @@
...
+
+## Gather Tools
+
+- [Backups](/articles/backup-files)
+- [Live Disk](/articles/live-disk)
+- [Logs](/articles/create-logs)
+
+## Suspend, Reboot, Shut Down
+
+Do the symptoms go away if the system is suspended, rebooted or shut down and then started up again? Sometimes repeated suspends or reboots are all that are needed to clear an issue out. It's also a good idea to test reboot and shutdown separately as they handle memory and power-management differently.
```
leviport, line 28: Should we recommend suspend/resume cycles for troubleshooting? In my experience, suspend is usually the cause of problems, not the solution.

storyaddict: Suspend and resuming will sometimes fix gdm refresh issues if the display is stuck. Shut the lid, open it again, and we're back.

leviport: Switching TTYs seems like a better way to refresh gdm, but fair enough.

ahoneybun (changes requested):
I think we should get some of this if not most of it to the Ubuntu article as well.

content/package-manager-pop.md
```diff
@@ -54,6 +55,52 @@ sudo apt install <package name>
 sudo apt install pop-desktop
 ```
 
+### 4. Update with Packages Held
```
ahoneybun, line 58: We should remove this number.

ahoneybun (changes requested):
We can move most of this to the Ubuntu article and maybe replace flatpak with snap.

gangwerz (approved):
Looks Good

PR #763, opened 2021-08-23
Test i-tec USB-C Dual Display Docking Station

pspacek:


jacobgkau (approved):
Thank you for the compatibility report!

PR #766, opened 2021-08-28
Fix the lsblk command

jbrains:
I assume that this fixes the lsblk command, because I'm not sure what it could have done as it was written.


jbrains:
I found a second problem, so I put both in https://github.com/system76/docs/pull/767

PR #767, opened 2021-08-28
Fix log files commands

jbrains:


jacobgkau (approved):
Thank you for the corrections! (cc: @ahoneybun, @thomas-zimmerman)

PR #768, opened 2021-08-30
update firmware keys and add a link to the Open Firmware article

ahoneybun:
This PR does the following:
- Adds the updated table for booting keys from the live-disk article to the pop-live-disk article
- Updated the live-disk and pop-live-disk articles to have a link to the Open Firmware system article

This clears up some confusion with which keys to use to boot to the BIOS and Boot Menu for Open Firmware systems.


storyaddict (approved):
changes look good

PR #770, opened 2021-09-02
move article to another slot

ahoneybun:
This PR does the following:
- Moves the Clear cache and cookies article to software troubleshooting


Socialhappiness (approved):
thank you for making the change. Approved

PR #772, opened 2021-09-07
updating software triage

Socialhappiness:
Removed 'Causes.' There's too much negative verbiage in there and it's not necessary. You have symptoms and you have solutions.
Are all the symptoms necessary? Can they be described without negative words? 
Changed suspend-reboot-shutdown section
Adding suspend and reboot in the beginning doesn't make sense to me and might not to the customer. It overcomplicates the first suggestion to try turning it off and on again. 
Removed Solutions section: causes unnecessary scrolling and is unnecessary if there is a table of contents to the side. Delays getting to the actual solution. The verbiage was fun, but after more thought, it's unnecessary.
The purpose of this article is to get the customer to a solution when they are having software issues.


content/software-triage.md
```diff
@@ -23,9 +23,9 @@ tableOfContents: true
 - [Live Disk](/articles/live-disk)
 - [Logs](/articles/log-files)
 
-## Suspend, Reboot, Shut Down
+## Shut Down
 
-Do the symptoms go away if the system is suspended, rebooted or shut down and then started up again? Sometimes repeated suspends or reboots are all that are needed to clear an issue out. It's also a good idea to test reboot and shutdown separately as they handle memory and power-management differently.
+Sometimes turning the machine off and on again is all that's needed to clear an issue out. Shut down your computer and power it back on to see if the issue persists.
```
storyaddict, line 28: There can be markedly different behavior between shutdowns, reboots and suspends. I specified these out for a reason. :)

content/software-triage.md
```diff
@@ -34,7 +34,7 @@ Do the symptoms go away if the system is suspended, rebooted or shut down and th
     Symptoms like those described [here](/articles/login-loop-pop) or [here](/articles/login-loop-ubuntu)
 - Applications will not open.
     Clicking on the application icon in the Dock does not launch the program, and/or entering the package name in the terminal will not launch the program.
-- Updates will not complete.
+- Updates will not complete. Error messages show in the Pop! Shop or terminal output with the following;
```
storyaddict, line 37: Should be a colon not a semicolon I think.

I like this addition though.

content/software-triage.md
```diff
@@ -46,44 +46,8 @@ Do the symptoms go away if the system is suspended, rebooted or shut down and th
     E.g. windows frozen in place, windows will not resize. Stuck in full-screen mode, etc.
     Mouse freezes intermittently or stutters it moves across the screen, or keyboard input lags.
 
-## Causes
```
storyaddict, line 49: These are not negative words, they are descriptive words. And they are all potential causes of software issues.

The software is corrupted because the code was written correctly but something happened to it in download or install.

Old software is out of date.

If the wrong driver or driver version is installed, its an incorrect version.

content/software-triage.md
```diff
@@ -46,44 +46,8 @@ Do the symptoms go away if the system is suspended, rebooted or shut down and th
...
-    Many applications refer to system libraries. Some programs may be expecting to find a specific version of a library.
-- Out of date drivers.
-    Older kernels may be using out of date driver software.
-- Incorrect Drivers.
-    If an unneeded or incorrect driver is installed this can cause unexpected side effects.
-- Kernel Bugs
-
 ## Solutions
 
-We troubleshoot issues with the approach of taking small steps, then gradually larger steps:
```
storyaddict, line 65: I think its important for customers to see the logic in how we're troubleshooting so that when we follow these same procedures on the ticket they're already primed for our way of thinking.

storyaddict (changes requested):
Made some counterpoints. Maybe we can change the article slightly for a middle-ground?

PR #773, opened 2021-09-09
Test Dell WD19S Docking Station

pspacek:


pspacek:
Dock WD19 is already in the list, but it is sold out and replaced by WD19S.

jacobgkau (approved):
Thank you for the compatibility report.

pspacek:
Oh, sorry for removing the old model. My copy&paste-fu failed.

PR #776, opened 2021-09-15
update to remove the word disaster

Socialhappiness:
It's Data Recovery. It's not disaster recovery :) 
The link should be changed as well.


storyaddict (approved):
I think Disaster is useful for keyword searching but I'm not married to the title.

PR #778, opened 2021-09-22
Change year in text

bobodrone:
Under the section "Refresh Install" I have changed the release 19.04 => 21.04.

"Starting with new installations of Pop!\_OS 21.04"


jacobgkau (changes requested):
I'm not sure this change is appropriate without further rewording of the sentence; the point was that the "refresh install" feature was first available in Pop!_OS 19.04 ("starting with"), and remains available in all versions after that. This change makes it sound like it's not available for 20.04 LTS, for example, which is incorrect.

I'm all for rewording the entire sentence-- we shouldn't be using the future tense to refer to a current feature like that. For the bare minimum change, we could just remove everything up through the word "new," and have the sentence start with "Refresh Install is an option that allows you to..."

bobodrone:
I agree, I actually misread the text (non native English speaker). :) The first line "Refresh Installs are only available on a fresh install of Pop!_OS 19.04 and above." makes sense to have the year in the past, but actually I read the last paragraph is if it was an instruction to start using 21.04 with a fresh new install. My bad. 

But I agree, in the second case there is no need for repeating the year again.

storyaddict:
This is being addressed in this PR: https://github.com/system76/docs/pull/815 thanks @bobodrone !

PR #779, opened 2021-09-25
update pop shell for Manjaro

perkinsms:
Pop Shell no longer installed by default for Manjaro but according to this is available in the repository:

https://forum.manjaro.org/t/pop-shell-toggle-in-layouts-doesnt-work/68182/5


jacobgkau (approved):
Confirmed pop-shell is no longer installed by default (looks like it was removed when it wasn't working with GNOME 40) and that the installation instructions work on a Manjaro installation. `pacman` would also work since this is a binary package in Manjaro's repository, but `pamac` appears to be Manjaro's primary package management command, so this is fine for documentation.

PR #781, opened 2021-09-28
Fix command typo

Schykle:
Changed nnvidia to nvidia


ahoneybun:
That's not actually a typo and instead it was a change in apt 2.0. It is part of the new patterns I believe:

https://lists.debian.org/debian-devel/2020/03/msg00052.html

Schykle:
> That's not actually a typo and instead it was a change in apt 2.0. It is part of the new patterns I believe:
> 
> https://lists.debian.org/debian-devel/2020/03/msg00052.html

Ah I see! Totally missed that 😁 Thank you!

PR #785, opened 2021-09-30
Table-of-contents-fixed

storyaddict:
This PR makes sure that the table of contents option is present in the header for all current support docs.


thomas-zimmerman (approved):
LGTM

PR #787, opened 2021-10-01
Add an AMD FSR Guide

Serious-Table:
I've written a short guide about using AMD FSR with games running in Steam with Proton-GE. There is also a mention this will work in Lutris when 0.5.9 is out of beta, but presently does not.

I've included a single image of recommended resolutions, taken from an AMD presentation.


storyaddict (approved):
Very well written article addition. Thank you @Serious-Table for the addition! :)

PR #790, opened 2021-10-05
Updates to include plugable ud-ultc4k

jtmackoy:
After [updating firmware to latest](https://github.com/system76/firmware-open/commit/14b8a6e967db28596d7108b7a925c60c416d7f45) on my gaze16_3060, I am now able to successfully use the Plugable ud-ultc4k for my needs, which include:

- Connecting two external displays via the DP++ ports on the dock - one display uses a DP -> HDMI adapter and the other uses a DP -> DVI adapter.
_Note: my displays are not pushing 4K resolutions - both max out at 1920 x 1080_
- Connect multiple USB devices (using all of the USB-A ports - haven't tested the USB-C port on front of dock):
  - USB headset
  - Logitech USB Unifying Receiver
  - Logitech USB Webcam
 - Connect Ethernet (confirmed link at 1Gbps)


jacobgkau:
Thank you for the compatibility report! I'm going to make a couple of small tweaks:
- Moving `UD-ULTC4K` above `UD-ULTCDL` in the list since numbers come first in alphabetical order.
- Revising the name of both the `UD-ULTC4K` and the `UD-ULTCDL` so they're more consistent with one another, as compared with the product pages they're linking to.
- Merging the note about the `gaze16` graphics requirement with the existing `gaze15` note.
- Moving the note about graphical issues in NVIDIA mode under the dock itself; the superscript notes are for items that affect more than one dock.

jacobgkau (approved):

PR #797, opened 2021-10-08
Update docking-station.md

bergluc:
Add the Anker dock

### DIFF
--- a/content/docking-station.md
+++ b/content/docking-station.md
@@ -56,7 +56,7 @@ We have tested the following docks:
 
 Community members have reported that the following docks work with our products:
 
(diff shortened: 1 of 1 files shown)


jacobgkau:
Added the link and removed the following notes from this user:
> Tested with a Lemur Pro. It worked well to connect a keyboard and mouse (USB), Ethernet cable, speakers (audio out) and monitor (DisplayPort). The docking station should be turned off initially and then turned on before starting up the laptop to be sure the monitor is recognized. Otherwise, you will need to open the laptop cover to turn on the monitor.

rpsirois:
Thank you for all the work testing this dock. I recently purchased a Lemur Pro with the Intel Iris Xe graphics and ran two Asus Pro Art monitors via HDMI through the dock (QHD 1440 resolution). I have been having intermittent issues with the monitors going in and out. I can always seem to run one just fine, but sometimes the second one won't get a signal. Thoughts?

PR #810, opened 2021-10-29
Fixing a typo

1105420698:
it should be /usr/share/xsessions, notice the sessions word has a s.


leviport (approved):
Thank you for the correction!

Issue #809, opened 2021-10-29
Deprecate the "Install Tensorflow" article


gangwerz:
The Tensorflow package that is the centerpiece of this article is no longer in the repositories. The officially recommended way to use Tensorflow is to use containers. The Tensorman article is preferred in those cases, at it makes the container process easier.

https://support.system76.com/articles/install-tensorflow

PR #816, opened 2021-11-03
removing tensorflow article tensorman preferred

storyaddict:
This PR addresses issue https://github.com/system76/docs/issues/809 by removing the tensorflow article. 

The version of tensorflow referenced in that article is no longer in the repositories. Deferring to tensorman, containerized approach.


jacobgkau:
I'm going to add a redirect on the Tensorman article so anyone who had this one bookmarked gets to the new article instead of a 404 page.

storyaddict:
> ~I'm going to add a redirect on the Tensorman article so anyone who had this one bookmarked gets to the new article instead of a 404 page.~ The redirect was already set (not sure how that works)... the "Use TensorFlow" article should also be removed, and also already has its redirect set in the Tensorman article.

I apologize if I'm being dense...the purpose of this PR is to remove the Tensorflow article. Do I need to do something else first?

ahoneybun (approved):
This PR does remove the TensorFlow articles and leaves the Tensorman article alone.

Issue #674, opened 2021-06-17
Add xorg.conf reset as suggestion for solving "black screen" problem


josephrocca:
I had a problem with my adderws laptop screen being black/blank when I booted up, and I went through the trouble-shooting steps [listed here](https://support.system76.com/articles/login-loop-pop/). None of them worked, but I eventually solved it thanks to [this](https://askubuntu.com/questions/1059965/internal-laptop-screen-not-detected-when-using-nvidia-driver) AskUbuntu post. The solution was just to delete the `xorg.conf` file:
```
mv /etc/X11/xorg.conf /etc/X11/xorg.conf.old
```
Might be worth adding to the aforementioned docs article?

storyaddict:
@josephrocca thanks for bringing this our attention. I think this is a good addition to include.

PR #818, opened 2021-11-04
Update black screen

storyaddict:
This PR addresses suggestion made in issue: https://github.com/system76/docs/issues/674 

Specifically, it adds notes on backing up/removing xorg.conf as a possible troubleshooting solution for a black screen on login.

Both the Pop and Ubuntu articles have this addition.

Also removed Ubuntu specific instructions from Pop article.


storyaddict:
> Changes look good but one note would be do we want to link to the boot-loader article in case they can't get to a TTY?

I like that idea! I'm going to do that in a separate PR though since there's also an issue that refers to updating the bootloader article.

Issue #805, opened 2021-10-15
Pop!_OS Refresh 


gangwerz:
A customer mentioned in a ticket that they were unsure if they should click "Refresh" while updating Pop Recovery when they were going through the `pop-upgrade` article. 

I think mentioning the "Refresh" option in this article, and explaining that they shouldn't use that now.

A link to an article that covers how to use the "Refresh" feature would also be useful.

PR #819, opened 2021-11-04
Update pop upgrade article

storyaddict:
This PR does the following: 

- updates and prepares the Pop Upgrade article for the upcoming release of 21.10 (Exact date TBD).
- updates formatting for consistency with other articles.
- Adds reassuring preamble back to the article that was cut in most recent revision.
- Address this issue: https://github.com/system76/docs/issues/805 by explaining *not* to use the Refresh OS option during upgrades.


content/upgrade-pop.md
```diff
@@ -74,27 +85,31 @@ sudo apt update
...
 
-Once the updates are applied, a notification should appear at the top of your screen saying that an upgrade is available. Click on this notification, or go to Settings ->  OS Upgrade & Recovery. The System76 upgrade package will display a message that Pop!\_OS 21.04 is available with a **Download** button.
+Once the updates are applied, a notification should appear at the top of your screen saying that an upgrade is available. Click on this notification, or go to Settings ->  OS Upgrade & Recovery. The System76 upgrade package will display a message that Pop!\_OS 21.04 is available with a `Download` button.
+
+If you are planning on staying on an LTS release for the time being, this is also the page where you can dismiss upgrade notifications.
 
 ![Settings OS Upgrade](/images/upgrade-pop/setting-os-upgrade.png)
 
-1. Click the **Update** button to update the Recovery partition.
+**IMPORTANT NOTE:** Clicking `Refresh` to the right of "Refresh OS" will reinstall the Operating System with the version stored in Recovery. Refreshing is **not** part of the upgrade process.
```
jacobgkau, line 94: Does this need to be worded so strongly given that it's basically saying "you don't need to click this button that we don't tell you to click?"

storyaddict: Looking at this again, I agree with you. However, Zach's issue was referring to customers clicking update, update Recovery, and Refresh (without us telling them to).

I can soften/clarify this though.

storyaddict: Changes have been made.

jacobgkau: The issue was referring to a customer who was following along with the article, and it sounds like they were overthinking things if they had to ask about an option that the article they were reading didn't mention.

If customers _not_ reading the article think they have to refresh, that sounds like more of a UX issue than something that should be worked around in documentation.

I think this article was previously talking about the Recovery partition because you wanted to make sure it was updated in case the upgrade went wrong? pop-upgrade now upgrades Recovery automatically before performing a release upgrade, so this entire step may no longer be necessary: https://github.com/pop-os/upgrade/pull/211/commits/0ea0f4de54e417bc47266a2b22e27ae0dc800d43

storyaddict: I think the issue is that "Refresh" and "Upgrade" have similar connotations mentally. Users will often mix up terms like this.

The UX could be improved...but I think having a clarifying note is a good idea in the meantime. Maybe the Refresh install option should be segmented out into its own block? Or more details printed about erasing installed apps.

I'll take out the step about upgrading recovery since that's now a "package deal."

gangwerz: Yeah, we are trying to address customer's overthinking things. But, they commented that "Refresh" sounds like something they should do to verify they are fully up to date after an update could lead to some very.... problematic tickets

gangwerz (approved):
I like these changes, and am not noticing any typos

storyaddict:
NOTE: even with an approving review, we won't be merging this without managerial sign-off since it mentions 21.10 as if it's out.

storyaddict:
@thomas-zimmerman @leviport please let me know when I can un-comment the Pop 21.10 section and we can merge this.

Issue #806, opened 2021-10-19
TLC for the Keyboard Shortcut Articles


gangwerz:
The Keyboard Shortcut articles could use a refresh. At first glance:

- The ubuntu-keyboard-shortcuts.md does not have the same explanation of the `SUPER` key. A mention that it means the same as the `Windows` key would be good
- If we could make a PDF of this page available for download, I know some customers would appreciate the offline cheat sheet

PR #822, opened 2021-11-05
update icons and text in keyboard shortcuts

ahoneybun:
This PR does the following:
- Remove the Pop icon as it is confusing to some folks
- Updating text so that both articles match a bit better
- Adds the pop key icon svg
- Address this issue: https://github.com/system76/docs/issues/806


storyaddict:
I think we should decide what the convention is going to be for referencing the SUPER key, then use this article as the example to revise others.

storyaddict (changes requested):
Please add SUPER key icon to Ubuntu article as well.

storyaddict (approved):
I made the changes requested of Aaron.

Issue #783, opened 2021-09-30
Update Desktop environment change article.


Atharva-Phatak:
Hi, 
I recently changed installed another desktop environment from pop_os default to Kde-plasma. Unfortunately, KDE-plasma messed up the pop_os UI. 
I tried to remove it but yeah something was messed up and I did a clean install. 
What I am proposing is that we add a few warnings/Notes to the page that additional desktop environment could hamper the default pop_os settings ?

Just  a recommendation. :)

PR #823, opened 2021-11-05
add note about changing desktops

ahoneybun:
This PR does the following:
- Adds a note about changing desktops
- Addresses this issue: https://github.com/system76/docs/issues/783


storyaddict (approved):
Formatting and calling attention to the "Note" looks good.

PR #829, opened 2021-11-09
Linux Gaming: Prioritize pop shop for installing steam

noahtallen:
Since these articles are aimed at users who may be less comfortable with Linux, I think it makes sense to display Pop Shop as the first item, rather than terminal. I think for new users, the terminal is likely more of a fallback option, rather than their primary means of installing software.

This PR was motivated by this comment on Reddit from @mmstick: https://www.reddit.com/r/pop_os/comments/qqa44e/linus_tech_tips_switches_his_personal_pc_to_pop_os/hjz960n/

### DIFF
--- a/content/linux-gaming.md
+++ b/content/linux-gaming.md
@@ -20,6 +20,14 @@ tableOfContents: true
 
 ## Steam (Install)
 
+### Install Steam From the Pop!_Shop
+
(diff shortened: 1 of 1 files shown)


jacobgkau (approved):
I see no reason why the terminal instructions should be first in this article, and looking back through the article's history, it doesn't appear to have been a deliberate decision. Thank you for the tweak.

jacobgkau:
We may want to make the same change for the Gamehub and Lutris sections.

PR #832, opened 2021-11-11
Rename "Using Geary" to "Set Up E-Mail"

gangwerz:
Renamed the email article to include the word "email" so it's clearer what the article is about.


storyaddict (approved):
Title change makes sense. Formatting inconsistency in description is intentional.

Issue #765, opened 2021-08-27
nvcc --version not working after following CUDA cuDNN install article


John15321:
I have been following https://support.system76.com/articles/cuda article (I also installed Pop OS with).

I did:
```bash
sudo apt install system76-cuda-latest
```

and

```bash
sudo apt install system76-cudnn-11.2
```

so that it matched my cuda version (it installed 11.2)

But now I have 2 problems.

1. `nvcc` command is not working so Im guessing something went wrong
2. When I run `nvidia-smi` I get this:
```bash
Fri Aug 27 23:03:58 2021       
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 470.57.02    Driver Version: 470.57.02    CUDA Version: 11.4     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  NVIDIA GeForce ...  Off  | 00000000:01:00.0  On |                  N/A |
| N/A   42C    P0    23W /  N/A |    878MiB /  5936MiB |     17%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
| Processes:                                                                  |
|  GPU   GI   CI        PID   Type   Process name                  GPU Memory |
|        ID   ID                                                   Usage      |
|=============================================================================|
|    0   N/A  N/A      1704      G   /usr/lib/xorg/Xorg                305MiB |
|    0   N/A  N/A      1956      G   /usr/bin/gnome-shell              275MiB |
|    0   N/A  N/A      8519      G   ...AAAAAAAAA= --shared-files      183MiB |
|    0   N/A  N/A     15817      G   .../debug.log --shared-files       24MiB |
+-----------------------------------------------------------------------------+
```

Which would indicate I have CUDA 11.4 so wtf pls help

I need CUDA and cuDNN for machine learning

My system (output from `inxi -Fxxx`):
```bash
System:    Host: pop-os Kernel: 5.11.0-7620-generic x86_64 bits: 64 compiler: gcc v: 10.2.1 Desktop: GNOME 3.38.4 
[...]

PR #837, opened 2021-11-12
Update cuda cudnn

ahoneybun:
This PR does the following:
- Removes a `cat` command that pointed to a file that doesn't exit
- Adds a new `cat` command to replace it to check cuDNN version instead
- Adds and removes CUDA/cuDNN versions as they are no longer packaged
- Adds information that certain CUDA/cuDNN versions are only on 20.04 LTS
- Addresses issue https://github.com/system76/docs/issues/765

### DIFF
--- a/content/cuda.md
+++ b/content/cuda.md
@@ -32,6 +32,12 @@ sudo apt install system76-cudnn-11.2
 
 ### For older releases of The NVIDIA CUDA Toolkit
 
+To install CUDA 11.1
+
+```bash
+sudo apt install system76-cuda-11.1
+```
+
 To install CUDA 10.0:
 
 ```bash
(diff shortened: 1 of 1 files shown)


ahoneybun:
I checked the following:

Pop 20.04 LTS:
- CUDA and cuDNN 10.0/10.1/10.2

Pop 21.04:
-  CUDA and cuDNN 11.1/11.2

ahoneybun:
I checked the following:

Pop 20.04 LTS:
- CUDA and cuDNN 10.0/10.1/10.2

Pop 21.04:
-  CUDA and cuDNN 11.1/11.2

storyaddict (approved):
Changes look good, and version numbers have been double-checked.

PR #838, opened 2021-11-14
Update doc to mention install of ffmpeg

mattmcgiv:
I ran through this setup this morning and was able to get howdy working. I just wanted to update the doc to include a mention of the need to install ffmpeg. When I originally tried to run ffplay I got an error saying that I should try installing ffmpeg first. Hope it helps, thanks!

### DIFF
--- a/content/setup-face-recognition.md
+++ b/content/setup-face-recognition.md
@@ -98,6 +98,12 @@ ffplay /dev/INPUT
 
 Replace `INPUT` with results you had from previous command.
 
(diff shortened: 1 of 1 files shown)


storyaddict (approved):
Change looks correct format-wise.

jacobgkau (approved):
The article directs the user to run `ffplay` (to test camera inputs by path), which is provided by the `ffmpeg` package. The `ffmpeg` package is not installed by default. This change is accurate.

PR #840, opened 2021-11-15
Update pop-recovery.md

linuxgnuru:
added command line for upgrading the recovery partition


ahoneybun (approved):
Changes look good! :tada:

PR #841, opened 2021-11-19
Changed text for machine checks

linuxpng:
I think that there is a typo in the memory section regarding corrected and uncorrected errors.  Customers without ECC RAM should never see corrected errors in their logs.  I also think you are intending to say that if you have ECC RAM and there are a large number of corrections in a short period of time it may still indicate a fault.

### DIFF
--- a/content/hardware-failure.md
+++ b/content/hardware-failure.md
@@ -178,7 +178,7 @@ Then, after the system has crashed or been used for a period of time, take a loo
 journalctl -f -u rasdaemon
 ```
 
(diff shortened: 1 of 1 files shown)


jacobgkau:
Just a heads-up, the old reading sounded to me more like it was referring to pcieport AER `severity=Corrected` log lines, which it looks like rasdaemon can also report about. Support team may want to determine how those logs and the rasdaemon output is used when reviewing this change.

linuxpng:
Ah, that's a good point.  Maybe a section for both PCI and RAM errors?

PR #842, opened 2021-11-23
add oryp8 and pang11

ahoneybun:
This PR does the following:
- oryp8
- pang11


gangwerz (approved):
Changes look good, no typos from what I can see

PR #844, opened 2021-11-24
Launch keyboard configurator article

ahoneybun:
This PR does the following:
- Adds a new article about the Launch keyboard


content/launch-keyboard.md
```diff
@@ -0,0 +1,68 @@
...
+
+## The Launch and Accessories
+
+The following items are included with your Launch:
+
+![Accessories](/files/launch-keyboard/launch-accessories.webp)
+
+## QMK firmware
+
+The Launch keyboard uses our version [QMK firmware](https://github.com/system76/qmk_firmware) and it is built using our changes that are not in upstream [QMK firmware](https://github.com/qmk/qmk_firmware). You can read more about QMK [here](https://github.com/system76/qmk_firmware).
```
jacobgkau, line 26: > The Launch keyboard uses our version [QMK firmware](https://github.com/system76/qmk_firmware)

Do you mean "our version _of_ QMK firmware"?

content/launch-keyboard.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Layers
+
+The core idea of "Layers" is that one key can be used in different setups for example on our laptops holding the <kbd>Fn</kbd> then pressing the <kbd>F5</kbd> will lower the volume while pressing <kbd>F5</kbd> on it's own will perform a different action like refreshing a web page if you are in a web browser like <u>Firefox</u>. You can read more about layers [here](https://beta.docs.qmk.fm/using-qmk/software-features/feature_layers
```
jacobgkau, line 30: > one key can be used in different setups

"different setups" is vague and probably not the best way to word this.

> The core idea of "Layers" is that one key can be used in different setups for example on our laptops holding the <kbd>Fn</kbd> then pressing the <kbd>F5</kbd> will lower the volume while pressing <kbd>F5</kbd> on it's own will perform a different action like refreshing a web page if you are in a web browser like <u>Firefox</u>.

This is a run-on sentence. I would recommend splitting the sentence before "for example" (and add the appropriate commas in the second sentence.)

> You can read more about layers [here](https://beta.docs.qmk.fm/using-qmk/software-features/feature_layers

The closing parenthesis for this link probably shouldn't be on a separate line.

content/launch-keyboard.md
```diff
@@ -0,0 +1,68 @@
...
+).
+
+## Firmware Updates
+
+Firmware updates for the Launch are handled using [fwupd](https://fwupd.org/) and our <u>Firmware Manager</u> can be used to apply the update. When you get a notification about a firmware update you will need to unlock the device using Fn+Esc (note that this will be the default keys even if you remapped them in the Keyboard Configurator) before you can apply the update. When the keyboard is unlocked you will see the following keys light up:
```
jacobgkau, line 35: > When you get a notification about a firmware update

A comma is required after this dependent clause.

> When the keyboard is unlocked

A comma is required after this dependent clause.

content/launch-keyboard.md
```diff
@@ -0,0 +1,68 @@
...
+
+- U N L O C K E D
+
+Once the keyboard is in this mode you can apply the firmware update, when the update is complete you will need to unplug and re-plug it in to exit this mode.
```
jacobgkau, line 39: > Once the keyboard is in this mode

A comma is required after this dependent clause.

> you can apply the firmware update,

This needs to be a period or a semicolon, not a comma.

> when the update is complete

A comma is required after this dependent clause.

> you will need to unplug and re-plug it in

"re-plug it in" sounds a little awkward. It might be worth rewording.

content/launch-keyboard.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Keyboard Configurator
+
+You can download the `AppImage` for Ubuntu and other OSes and use the <u>Pop!\_Shop</u> for Pop!\_OS. Use this [link](https://github.com/pop-os/keyboard-configurator/releases
```
jacobgkau, line 43: > use the <u>Pop!\_Shop</u> for Pop!\_OS

Might want to change this to "on Pop!_OS."

The closing parenthesis for the link should probably not be on a separate line.

content/launch-keyboard.md
```diff
@@ -0,0 +1,68 @@
...
+) to download the application.
+
+You can also download it from the Arch Linux AUR. The main version is built from latest releases which is recommended for most users which you can find [here](https://aur.archlinux.org/packages/keyboard-configurator/
```
jacobgkau, line 46: > You can also download it from the Arch Linux AUR. The main version is built from latest releases which is recommended for most users which you can find [here]

This is a run-on clause. Each "which" phrase should have a comma before it, and it would be better if there weren't two in a row.

content/launch-keyboard.md
```diff
@@ -0,0 +1,68 @@
...
+) and the git version is built from the latest commits to the project (this version will have the very latest fixes and features) which can be found [here](https://aur.archlinux.org/packages/keyboard-configurator-git/
+).
```
jacobgkau, line 48: Both of the links in this paragraph should have their closing parenthesis on the same line.

content/launch-keyboard.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Remapping Keys
+
+<video width="900" height="900" controls>
+  <source src="/files/launch-keyboard/remapping-function-keys.webm" type="video/mp4">
+</video>
+
+The keys are defined in the firmware of your keyboard and you can remap the keys using the <u>Keyboard Configurator</u>. After opening the <u>Keyboard Configurator</u>, you can then select the key you would like to configure, and select an item from the available items below the keyboard showing in the application. I am attaching a video showing the remapping process. Please note the "Layer 1" and "Layer 2" at the top of the configurator. Layer 1 is what the keys will do normally, then Layer 2 shows what the keys do when "Fn" is held down.
```
jacobgkau, line 56: `you can remap the keys using the <u>Keyboard Configurator</u>` is an independent clause, so add a comma before "and" in the first sentence.

> you can then select

The word `then` is not needed here. It may make sense if this was the beginning of the sentence, but it's not.

> select the key you would like to configure, and select an item

I don't think there should be a comma before this "and". You're listing two things, the second section isn't functioning as an independent clause.

> I am attaching a video showing the remapping process.

You should not be referring to yourself in the first person. Something like `See the attached video showing the remapping process.` would be an alternative.

> Layer 1 is what the keys will do normally, then Layer 2 shows what the keys do when "Fn" is held down.

There is no time component to these buttons, so I would recommend using "and" or another word instead of "then."

content/launch-keyboard.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Changing LEDs
+
+<video width="900" height="900" controls>
+  <source src="/files/launch-keyboard/video-soothing.mp4" type="video/mp4">
+</video>
+
+In the <u>Keyboard Configurator</u> you can change the LEDs in the 'LEDs' section of the application by changing the 'Layer Color Pattern' using the dropdown.
```
jacobgkau, line 64: > In the <u>Keyboard Configurator</u>

A comma is required after this dependent clause.

content/launch-keyboard.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Tech Docs
+
+We have documentation that goes over changing the keycaps, switches and more in the tech docs which can be found [here](https://tech-docs.system76.com/models/launch_1/repairs.html).
```
jacobgkau, line 68: > We have documentation that goes over changing the keycaps, switches and more in the tech docs which can be found [here]

A comma is required before "which." I would also suggest an optional Oxford comma after "switches."

jacobgkau (dismissed):

content/launch-keyboard.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Remapping Keys
+
+<iframe width="900" height="500" src="/files/launch-keyboard/remapping-function-keys.webm" title="Remapping keys" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
+
+The keys are defined in the firmware of your keyboard, and you can remap the keys using the <u>Keyboard Configurator</u>. After opening the <u>Keyboard Configurator</u>, you can select the key you would like to configure and select an item from the available items below the keyboard showing in the application. See the attached video showing the remapping process. Please note the "Layer 1" and "Layer 2" at the top of the configurator. Layer 1 is what the keys will do normally and Layer 2 shows what the keys do when "Fn" is held down.
```
storyaddict, line 51: "After opening the Keyboard Configurator, select the key you would like to re-map, then select an item from the available mapping options shown in the list below the keyboard diagram. Refer to the included video for more details."

content/launch-keyboard.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Changing LEDs
+
+<video width="900" height="900" controls>
+  <source src="/files/launch-keyboard/video-soothing.mp4" type="video/mp4">
+</video>
+
+In the <u>Keyboard Configurator</u>, you can change the LEDs in the 'LEDs' section of the application by changing the 'Layer Color Pattern' using the dropdown.
```
storyaddict, line 59: "...you can change keyboard back light color and behavior in the `LEDs` section of the application. 

Is `Layer Color Pattern` the entry for both changing color and light behavior? Can we clarify this?

storyaddict (changes requested):

content/launch-keyboard.md
ahoneybun: I added a bit more about the Layer Color Pattern section.

storyaddict (approved):
Changes look good. Aaron made some changes, and so did I.

PR #846, opened 2021-11-26
Update install-pop.md

LTTForumCustomizer:
Added NVIDIA ISO note at the bottom of guide.

### DIFF
--- a/content/install-pop.md
+++ b/content/install-pop.md
@@ -87,3 +87,5 @@ If you installed from the Intel/AMD ISO and your system has discrete NVIDIA grap
 ```
 sudo apt install system76-driver-nvidia
 ```
+
+If you used the ISO with NVIDIA drivers included, this step is not required. 


content/install-pop.md
```diff
@@ -87,3 +87,5 @@ If you installed from the Intel/AMD ISO and your system has discrete NVIDIA grap
 ```
 sudo apt install system76-driver-nvidia
 ```
+
```
storyaddict, line 90: Please format as:

**NOTE:** If you used the ISO with NVIDIA drivers included, this step is not required.


You may also need to rebase to make sure the branch is up to date. :)

storyaddict (changes requested):

leviport:
Wouldn't this note be redundant because of the sentence on line 85?

jacobgkau:
I can no longer edit this file because @LTTForumCustomizer has deleted their branch that this PR was coming from.

I don't agree that a note needs to be labeled as a note, but I do agree that this one isn't strictly necessary, although I don't think it would hurt either. In any case, I'm closing this since it can't be modified further; a new PR is welcome to add the additional clarification sentence if anyone else thinks it's needed.

PR #847, opened 2021-11-27
Update pop-shell.md

mmahlguest:
sudo -k to not persist user's credential otherwise make will fail

### DIFF
--- a/content/pop-shell.md
+++ b/content/pop-shell.md
@@ -18,7 +18,7 @@ tableOfContents: true
 In order to build Pop Shell from source, we'll first need to install the build dependencies and <u>git</u> to clone the repository:
 
 ```bash
-sudo apt install git node-typescript make
+sudo -k apt install git node-typescript make
 ```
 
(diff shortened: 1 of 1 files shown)


jacobgkau (changes requested):
Can you please clarify the purpose of this change? The `-k` option would make `sudo` require a password again next time it's used; the `make local-install` command (the only `make` command present in this readme) does not use `sudo`, so this change should have no effect.

mmahlguest:
Quite frankly I didn't test exactly with -k, sorry.

I did follow the steps in the readme and was always getting error from make local-install, then I found a post somewhere saying to not run make in a root session. So ran sudo only to get the package, bounced shell (exist/reopen terminal), and then make worked well and I thought I should share what I experienced. Will test again when I can, I didn't have time to uninstall and try again.

________________________________

jacobgkau:
I am closing this PR because the change does not make sense. If you see the issue you're describing again, please get a copy of the terminal output when `make` fails and report the issue at https://github.com/pop-os/shell.

PR #853, opened 2021-12-05
Added some Steam specific information

truefakts:
There wasn't any mention of how the Pop Dock makes launching applications with PRIME offload really simple. Furthermore, this is a handy trick that makes it so games launched from the client use those flags automatically. Lastly the command line options were added for folks that might not want to use the Steam client directly, but rather would use the Gnome application menu


content/graphics-switch-pop.md
```diff
@@ -63,7 +63,34 @@ GLX applications must be launched with this command to be rendered on the dGPU (
 __NV_PRIME_RENDER_OFFLOAD=1 __GLX_VENDOR_LIBRARY_NAME=nvidia <application>
 ```
 
-Hybrid graphics mode is available on 19.10 and later.
+Hybrid graphics mode is available on 19.10 and later. Applications can be launched using Hybrid graphics on a per-application basis and through inheritance of launch state from a game launcher.
+
+#### Hybrid Graphics Integration with Steam: GUI Method
+
+When Steam is favorited, can be launched using Discrete Graphics by right clicking on the Pop Dock and selecting the following option:
```
jacobgkau, line 70: This should not be necessary. Steam's `.desktop` file contains `PrefersNonDefaultGPU=true`, which should make it use the dedicated GPU by default in Hybrid mode.

jacobgkau (changes requested):
Reviewing this PR again, it has the following issues:

- The statement that "Applications can be launched using Hybrid graphics on a per-application basis and through inheritance of launch state from a game launcher." is redundant with the above "Applications will use the integrated GPU unless explicitly requested to use the discrete GPU."
- Steam's `.desktop` file is set to launch on the dGPU by default, so the context menu would not show that option. Instead, it shows the option to explicitly launch on the integrated GPU.
- We don't refer to this as the "Pop Dock," it's just the dock or COSMIC Dock.
- Several sentences (like line 70) are incomplete or otherwise grammatically incorrect.

Given that the changes are generally inaccurate, I'm going to go ahead and close this PR. I do think it would make sense to show the right-click context menu options (not only for Steam, but just in general) alongside the CLI options that are listed, but that's different enough from this that it would work just as well as a new PR. The part about setting launch options in the Steam preferences may work better in the Steam section of the [gaming article](https://support.system76.com/articles/linux-gaming).

Thank you for being involved with the Pop!_OS community!

PR #862, opened 2021-12-11
Split Arch and Fedora articles for readability

nickgerace:
## Description

I recently received a Thelio Major for work and have installed Fedora on it. While trying to determine what packages, software, etc. I needed to get the best experience possible, I found the support articles great, but overwhelming. Thus, I've split the articles by distro: Arch and Fedora.

Moreover, I made some small edits, including splitting DKMS and ACPI DKMS into two options, adding a `--refresh` flag to a `dnf` command, and more. I'd be happy to split this commit into two for reviewers too.

(My co-workers will likely be using Arch on their Thelio Majors as well, so there's some bonus points here for me :sweat_smile:)

## Commit Message

- Split Arch and Fedora articles for readability. Ensure that these
  articles link back to their parent and vice versa.
- Add refresh flag for post-COPR dnf command for Fedora driver install.
- Split DKMS and ACPI DKMS into two options with a disclaimer.
- Add missing system76-power package install for Fedora.

## Testing

Used volta-managed node 14 and nuxt js to read and test the articles locally on Fedora 35.


nickgerace:
Closing PR after brief discussion with @ahoneybun on the matter. This layout is not desired at this time.

PR #863, opened 2021-12-13
Update pop keyboard

ahoneybun:
This corrects a mistake that 'file filename' will search for files and instead it is 'find filename'.


storyaddict (approved):
Change is appropriate and needed.

PR #869, opened 2021-12-25
Add Quick Look previewers comparison

aklsh:
[Sushi](https://gitlab.gnome.org/GNOME/sushi) is a file previewer for the GNOME desktop environment, similar to macOS's [Quick Look](https://developer.apple.com/design/human-interface-guidelines/macos/system-capabilities/quick-look/).


storyaddict (approved):
Nice addition, thanks! :)

PR #870, opened 2021-12-28
Include nvme-cli

ChristopherRogers1991:
The `nvme` command referenced by the docs is provided by the `nvme-cli` pacakge:

```
kubuntu@kubuntu:~$ dpkg -S nvme | grep `which nvme`
nvme-cli: /usr/sbin/nvme
```


ahoneybun:
Thank you for the PR and it looks good!

ahoneybun (approved):
:tada:

PR #871, opened 2021-12-28
Add Flatpak repair command

adam-stambaugh:
Some issues that popup with flatpak require the repair command. It's a safe command to use and will resolve some issues for users that aren't very familiar with flatpak.


leviport (approved):
The repair command has been useful to me on multiple occasions as well. Thanks for the addition!

ahoneybun:
The change looks good live but is there a reason to have the `--user` option? If so should we not have `--system` as well?

jacobgkau:
> The change looks good live but is there a reason to have the `--user` option? If so should we not have `--system` as well?

The Pop!_Shop installs things with `--user`, so `--system` should have nothing that needs to be repaired unless the user has manually installed things with `--system`. It wouldn't hurt to list both, though. (`--system` would need sudo.)

ahoneybun (approved):

PR #875, opened 2021-12-31
Fix broken link to Mattermost

knorrium:


jacobgkau:
This would also be accomplished in https://github.com/system76/docs/pull/874.

knorrium:
Closing in favor of #874

PR #879, opened 2022-01-06
Added how to install deja dup via flatpak

linuxgnuru:
in case apt isn't working, offering a second option for installing deja dup backup


storyaddict (approved):
Good improvement. Approving.

PR #880, opened 2022-01-06
minor edits

linuxgnuru:
* added "rm -f" to the sudo rm /etc/apt/sources.list.d/pop-os-ppa.sources to avoid an error the customer might be concerned about
* added sudo in front of systemctl restart pop-upgrade
* added "bash -c "rm -rf" to the empty recovery partition
* added -H to the sudo gedit /etc/fstab


storyaddict (approved):
Confirmed changes have been tested with John and Thomas. Formatting looks good.

PR #883, opened 2022-01-10
Wrong Type of -- for msktutil, this command fails

stefan-ffr:
I've used this script and i get errors with the double dashes because they're not the same type on the website. I've changed it


ahoneybun:
@jacobgkau can you check if this does need to be changed?

jacobgkau (approved):
Confirmed that copy/pasting the commands (which still need to be modified after pasting to fit a user's local environment) was not working due to the hyphens being the wrong type. Copy/pasting the hyphens does work after the change.

Thank you for the fix!

PR #888, opened 2022-01-16
Fix hibernation docs

ChristopherRogers1991:
The guide as written doesn't quite work (though the screenshots help resolve discrepancies). This fixes the small errors, notes some of the benefits of hibernation, and suggests using `hibernate-then-suspend` for the best of both worlds.


content/enable-hibernation.md
```diff
@@ -112,16 +116,23 @@ However, these limitations aside, hibernation is an often requested feature for
 
 8. Add a resume target for the kernel:
 
-    Using the UUID retrieved from `fstab` run this command:
+    Reboot the computer into the normal OS partion.
```
ChristopherRogers1991, line 119: If you don't reboot at this point, you get this error:

```
recovery@recovery:~$ sudo kernelstub -a "resume=UUID=9d17f35b-e942-41b5-801d-3c1e741d482e"
kernelstub.Drive     : ERROR    Could not find a block device for the a partition. This is a critical error and we cannot continue.
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/kernelstub/drive.py", line 56, in __init__
    self.esp_fs = self.get_part_dev(self.esp_path)
  File "/usr/lib/python3/dist-packages/kernelstub/drive.py", line 94, in get_part_dev
    raise NoBlockDevError('Couldn\'t find the block device for %s' % path)
kernelstub.drive.NoBlockDevError: Couldn't find the block device for /boot/efi
```

content/enable-hibernation.md
```diff
@@ -29,7 +29,11 @@ Hibernation is not currently enabled on Pop!\_OS by default. There are several l
...
 
-However, these limitations aside, hibernation is an often requested feature for Pop!\_OS. The purpose of this article is to provide steps for enabling hibernation for any users that would like it.
+However, these limitations aside, hibernation is an often requested feature for Pop!\_OS, and there are some real benefits to it. For example, hibernation (when compared to always using suspend) can improve the longevity of your battery - full charge/discharge cycles can be hard on your battery and shorten its life. Since hibernation completely shuts off power to the device, it can prolong the lifetime of your battery. Additionally, there are some security benefits. While your computer is suspended, the decryption key for your data volume remains in memory, and your device is vulnerable to certain physical attacks that can extract the key from RAM. If the device is hibernated, and the hibernation swap space is encrypted, the drive must be decrypted to obtain the key, which would be impossible without already knowing the key (this is as secure as a normal, full shutdown).
```
ChristopherRogers1991, line 32: > and the hibernation swap space is encrypted

I believe this is correct, given I have to enter my password for the computer to resume after it has been hibernated (and I assume this is why the guide moves from a swap partition to a swap volume in the LVM setup), but it's probably worth someone confirm that this is true.

ChristopherRogers1991: Pretty sure https://wiki.archlinux.org/title/Dm-crypt/Swap_encryption#With_suspend-to-disk_support confirms it, but again, good to have someone verify security claims.

storyaddict:
Additional feedback from another customer: 

> Still bug in the advanced install in that it does not handle existing partitions within an encrypted LUKs partition.
> 
> Another "bug" is the documentation for resizing the LUKs partition.
> It should also include  way to find out what the UUID is once they are created.
> 
> The way I got it to work was to mount each partition in /mnt/[PartName] then  "ls -ltr /dev/disk/by-uuid" cross check that with disks and in my case dm-1 was /, dm-2 was /home
> 
> Then add /home to fstab with the UUID for dm-2

content/enable-hibernation.md
thomas-zimmerman: The change between hibernation and suspend changes the security of the system not at all. Pop!OS uses a encrypted swap at all times. Hibernation-than-suspend will still have any encryption keys in memory on suspend. Battery life is unlikely to be changed by hibernation unless you are only rarely using the device and ended up with many additional charging cycles--in which case you should be powering off the device anyway. Hibernation does increase your ware on NVMe/SSD drives that will ware out the hardware over just using suspend.

thomas-zimmerman:
Ran through all of the commands on a clean VM with 21.10 to make sure commands worked. Took some of the security improvements for hibernation and added to the top of the article and created new PR: https://github.com/system76/docs/pull/945

storyaddict:
#945 address this PR. :)

PR #889, opened 2022-01-18
Add info about rebooting and checking installation

John15321:
Add information about the necessity of reboot in order to check if the installation is valid
related issues: https://github.com/system76/docs/issues/765


storyaddict (approved):
Thanks @John15321 for this addition.

PR #890, opened 2022-01-20
Make the SMMStore article more generic

gangwerz:
I left the SMMStore information in the article. Then, restructured to allow this article to work for any firmware flash, but with info on the SMMStore if that should resurface. Also indicates to remove all drives if they are unable to reach the firmware boot menu.

### DIFF
--- a/content/open-firmware-from-usb.md
+++ b/content/open-firmware-from-usb.md
@@ -80,7 +80,7 @@ The highlighted files pictured below (the hash will change based on file and mod
 
 After the flash drive is ready, power off your system. Then, power back on your system and tap the Esc key while the system is turning on. This will take you to the firmware boot menu, where you should select the USB flash drive from the list of boot options.
 
-_If you are unable to access the firmware boot menu by pressing the Esc button while powering on the system. Then, shutdown the laptop and remove all NVMe M.2 drive(s) in the system using the step by step instructions for your system linked above_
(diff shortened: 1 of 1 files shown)


thomas-zimmerman (approved):
This makes this more usable

jackpot51 (changes requested):
This PR does not change the fact that the linked firmware images are specifically for clearing SMMStore, not for updating via USB. We would need new images that preserve the SMMStore for that.

thomas-zimmerman:
@jackpot51 We've been using this article as a general overview how updating the firmware when we provide a USB image file with specific versions of firmware.

jackpot51:
Also, if we did have links to firmware ZIP files here, who will be keeping those up to date?

thomas-zimmerman:
We were thinking of how to stage a page for each model that has prior releases of the firmware. As I've been able to build release USB images, support would be adding those on request.

jackpot51:
> We were thinking of how to stage a page for each model that has prior releases of the firmware. As I've been able to build release USB images, support would be adding those on request.

I would feel better if this page was copied instead of renamed, and all references to SMMstore clearing firmware were removed, including the links. That way the old page is still there for whoever needs to use these special firmware versions. They are not useful for newer firmwares since removing CMOS battery will do the same thing.

gangwerz:
Moving to dedicated file for the USB firmware flashing. Closing out this PR without merging.

PR #892, opened 2022-01-21
Fixing broken link

gangwerz:
Relative link using `articles/manage-repos-pop` lead to the compounded and non-functional link below:

https://support.system76.com/articles/pop-basics/articles/manage-repos-pop

~Replaced the relative link with an absolute link~

Aaron updated and corrected the relative link.


ahoneybun (approved):
LGTM :tada:

PR #894, opened 2022-01-25
Update system76-driver.md

uncivilgentleman:
Installing just `system76-driver` wasn't enough to make setting charge thresholds work.  I think the documentation could be improved more but this at least works so users can set charge thresholds from Fedora.


Seibz (approved):
Looks fine to me

PR #902, opened 2022-02-02
update to reboxing.md

thomas-zimmerman:
Add in Lithium battery label check for the re-boxing a latop. Include link to PDF for a replacement label.


ahoneybun (approved):
LGTM live :tada:

PR #903, opened 2022-02-02
Created Flash Firmware From USB Article

gangwerz:
This PR adds a dedicated article for preparing and booting a USB in order to flash firmware.

Originally written as a change to the SMMStore, but a dedicated article seemed more appropriate:
https://github.com/system76/docs/pull/890


storyaddict (approved):
These changes make sense. Having a second article independent of SMMStore is a good idea.
Article looks good, and I tested live and screenshots are showing up correctly.

Issue #887, opened 2022-01-14
Upgrade from 20.10 to 21.10


mikemhenry:
There doesn't seem to be any documentation on how to update 20.10 to 21.10.

The "Upgrading older releases" section only lists 

`Open a terminal and follow the next set of instructions to upgrade from Pop!_OS 18.10, 19.04, or 19.10.`

But on the main upgrade article, it says:

```
Upgrading Pop!_OS to 21.10 from 21.04

    NOTE: For all other operating system versions refer to the instructions for upgrading from an earlier release (https://support.system76.com/articles/upgrade-pop/#upgrading-older-releases)
```
And those instructions don't list 20.10

ahoneybun:
These instructions do need to be updated a bit but for 20.10 it is better to either use the recovery partition to refresh the OS or perform a clean install.

mikemhenry:
Cool, I'm working with support right now on the best path forward, but wanted to make a note here to make sure the docs get updated

Ampersandstorm:
Resolved in updates included in pull request #911

PR #910, opened 2022-02-08
updating older releases section with table

Ampersandstorm:
added a table in Upgrading Older Releases section, added instructions for 20.10 as listed in issue #887 convo


Ampersandstorm:
realized I did the links wrong

PR #912, opened 2022-02-08
Update for new install instructions.

13r0ck:
21.10+ has different requirements to install and run tensorman.

### DIFF
--- a/content/tensorman.md
+++ b/content/tensorman.md
@@ -29,10 +29,10 @@ redirect_from:
 sudo apt install tensorman
 ```
 
-For NVIDIA CUDA support, the following package must be installed:
+For NVIDIA CUDA support, the following packages must be installed:
 
 ```bash
-sudo apt install nvidia-container-runtime
+sudo apt install nvidia-docker2 nvidia-container-toolkit
 ```
 
 The user account working with Tensorman must be added to the `docker` group if that hasn't been done already:
@@ -41,7 +41,13 @@ The user account working with Tensorman must be added to the `docker` group if t
 sudo usermod -aG docker $USER
 ```
 
-If Docker was just installed, then a reboot will be needed before Tensorman can be used.
+THe last thing is to add a kernel parameter:
+
+```bash
+sudo kernelstub --add-options "systemd.unified_cgroup_hierarchy=0"
+```
+
+and reboot. Then you are ready for liftoff!
 
 ## About Tensorman
 
--- a/content/tensorman.md
+++ b/content/tensorman.md
@@ -29,11 +29,11 @@ redirect_from:
 sudo apt install tensorman
 ```
 
-For NVIDIA CUDA support, the following packages must be installed:
+For NVIDIA CUDA support, install the following packages, depending on your Pop!_OS version:
 
-```bash
-sudo apt install nvidia-docker2
-```
+| Pop!_OS 21.10 +                    | Pop!_OS 20.04 LTS                            |
+| ---------------------------------- | -------------------------------------------- |
+| `sudo apt install nvidia-docker2`  | `sudo apt install nvidia-container-runtime`  |
 
 The user account working with Tensorman must be added to the `docker` group if that hasn't been done already:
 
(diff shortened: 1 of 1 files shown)


content/tensorman.md
```diff
@@ -41,7 +41,13 @@ The user account working with Tensorman must be added to the `docker` group if t
 sudo usermod -aG docker $USER
 ```
 
-If Docker was just installed, then a reboot will be needed before Tensorman can be used.
+THe last thing is to add a kernel parameter:
```
jacobgkau, line 44: `THe` looks like an extra character is capitalized. I might also suggest "last step" instead of "last thing."

Might want to place an ellipses at the beginning of the sentence after the code block, since it's picking up the sentence from here.

13r0ck: Does it work now?

jacobgkau: I don't know why you went with "one last thing" instead of the suggested change. The problem was the ambiguity of the word "thing."

jacobgkau: I see that "Then you are ready for liftoff!" is being added by this PR too. Are you intentionally trying to introduce an unprofessional/casual voice?

13r0ck: > I don't know why you went with "one last thing" instead of the suggested change. The problem was the ambiguity of the word "thing."
Oh i just misread your message thats all. Either way its semantics at this point.

> I see that "Then you are ready for liftoff!" is being added by this PR too. Are you intentionally trying to introduce an unprofessional/casual voice?

:shrug: All of our messaging elsewhere is pretty casual, Is there a reason support-articles needs to be super professional?

jacobgkau: > Is there a reason support-articles needs to be super professional?

So people can take them seriously?

I moved the first one back closer to what you started with but replaced "thing" with "step," and contracted "you are" into "you're" in the final sentence since you're going for casual.

13r0ck: :+1:

content/tensorman.md
```diff
@@ -29,10 +29,10 @@ redirect_from:
 sudo apt install tensorman
 ```
 
-For NVIDIA CUDA support, the following package must be installed:
+For NVIDIA CUDA support, the following packages must be installed:
 
 ```bash
-sudo apt install nvidia-container-runtime
+sudo apt install nvidia-docker2 nvidia-container-toolkit
```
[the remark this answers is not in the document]
jacobgkau: Cool, I'll remove the second one just so there's less to type/copy.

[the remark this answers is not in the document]
13r0ck: :+1: Looks good to me

jacobgkau:
The instructions work on 21.10. On a 20.04 system, `nvidia-docker2` is not avaioable and `nvidia-container-runtime` is still the package to install. Do you think we should keep a section with the 20.04 instructions around as long as that release is still supported?

13r0ck:
@jacobgkau Probably. Knowing that 20.04 will only be LTS for ~2.5 months I'm not sure if that's totally necessary. I will leave the decision up to you about if we should add that & where/how it should be written

jacobgkau:
I know we generally don't recommend LTS over current, but we technically still support 20.04 for the next 2.5 months, so we should avoid breaking articles for it until then. I've added a table with the command for each release, based on how support's been using tables in other articles (such as the live disk/recovery articles.) Based on my testing, the kernel parameter isn't needed on 20.04, but also doesn't hurt.

jacobgkau (approved):
Someone else should take a look at the table to make sure that looks acceptable, but aside from that, this looks good to me. Gets the article working again on 21.10.

13r0ck:
Also just to note, the kernel parameter is only necessary on 21.10+

jacobgkau:
> As you said, the kernel parameter is only necessary on 21.10+. I'm not sure that is clear on your most recent PR. But if you are confident it doesn't matter then this should be good to go.

Yeah, I played with marking it as being for 21.10 only, but it seemed more complicated than it needed to be. To your point, 20.04 will be superseded very soon, so no harm having someone add the parameter now so it's there when they upgrade later. We're just listing the packages to avoid any show-stopping errors.

13r0ck:
> To your point, 20.04 will be superseded very soon, so no harm having someone add the parameter now so it's there when they upgrade later. We're just listing the packages to avoid any show-stopping errors.
Good point.

PR #915, opened 2022-02-09
Ending frozen applications

Ampersandstorm:
This articles lists a few methods to recover from completely frozen applications or desktop, based on personal recent experience. I haven't added the screenshot to /static/images yet, I'll need some help with that.


content/ending-frozen-applications.md
```diff
@@ -0,0 +1,76 @@
+---
+title: Ending Frozen Applications and Desktops
```
jacobgkau, line 2: I don't think this title is confusing at all, but I wonder if "Stopping" or "Closing" would be what users would search for first instead of "Ending".

content/ending-frozen-applications.md
```diff
@@ -0,0 +1,76 @@
...
+
+hidden: false
+section:
+---
+
+## Ending Frozen Processes, Applications, and Desktops
+
+Linux offers many methods to end hanging or frozen processes. The `gnome-system-monitor` provides a GUI process management application with features similar to the Windows Task Manager, while the `kill` command can end single or multiple processes from the Terminal.
+
+>Note: This article details several methods to end processes for a given scenario, but is not exhaustive or guaranteed to provide the most appropriate method for every scenario. Users should also keep in mind that suddenly terminating processes can result in unwanted data loss or system stability issues.
```
jacobgkau, line 24: > Note:

Can this `Note:` label be omitted? Is it unclear that this block of text is a note? As it relates to the second sentence, should the label say `Caution:` or `Warning:` instead?

> This article details several methods to end processes for a given scenario, but is not exhaustive or guaranteed to provide the most appropriate method for every scenario.

I'm not sure it's necessary or desirable to point out that the article doesn't cover all scenarios. That information isn't specific and may decrease confidence in the article.

> Users should also keep in mind that suddenly terminating processes can result in unwanted data loss or system stability issues.

This is a perfectly fair warning. Maybe this line could be reduced to just:

> Caution: Keep in mind that suddenly terminating processes can result in unwanted data loss or system stability issues.

content/ending-frozen-applications.md
```diff
@@ -0,0 +1,76 @@
...
+
+## End Specific Applications or Windows
+
+### Using gnome-system-monitor (GUI)
```
jacobgkau, line 28: This app shows up as `System Monitor` in the GUI menus/search results, so it might be better to refer to it as `System Monitor`. You can still provide the command to launch or clarify its technical name with parenthesis in the section body.

content/ending-frozen-applications.md
```diff
@@ -0,0 +1,76 @@
...
+
+The gnome-system-monitor comes prepackaged in Pop!_OS. This application provides a GUI interface and functionality similar to the Windows Task Manager.
```
jacobgkau, line 30: I would either omit "the" if you're using the command/app name as a proper noun, or add the word "app" after the name if you're referring to it that way.

content/ending-frozen-applications.md
```diff
@@ -0,0 +1,76 @@
...
+
+1. Press <kbd>Super</kbd> + <kbd>T</kbd> to launch a Terminal session.
```
jacobgkau, line 32: My instinct is to say that launching through the Launcher or App Library would be more user-friendly than opening a terminal for this particular app; it doesn't look like the terminal gives any useful output when running `gnome-system-monitor`. Did you encounter an instance where the GUI methods to launch System Monitor didn't work, but the GUI terminal still did work?

content/ending-frozen-applications.md
```diff
@@ -0,0 +1,76 @@
...
+2. Enter `gnome-system-monitor` and hit <kbd>Enter</kbd>.
+3. Locate frozen and hanging applications.
+
```
jacobgkau, line 35: This entire list is being rendered with excessive space between the list items due to these empty lines; leave the empty lines out to keep the list together. Compare how it looks now:
Screenshot from 2022-02-09 11-34-14
With how it should look:
Screenshot from 2022-02-09 11-33-57

content/ending-frozen-applications.md
```diff
@@ -0,0 +1,76 @@
...
+   ![gnome-system-monitor](/images/ending-frozen-applications/gnome-system-monitor)
+
+4. Right click the application, click Stop in the dropdown menu.
```
jacobgkau, line 38: Sounds like a word is missing after the comma; I would either add the word "then" after the comma, or remove the comma and use the word "and" instead, whichever you prefer.

Unrelated to the above, I wonder if the End (`SIGTERM`) or Kill (`SIGKILL`) options ever work when Stop (`SIGSTOP`) doesn't. I know there have been times when I needed to add a `-9` to a `kill` command for it to work, which is why I would think `Kill` might be needed in some cases. (Could be worth advising to try `Stop` first, then `End`, then `Kill` if you're trying to avoid having users jump straight to the most destructive option?)

content/ending-frozen-applications.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Using the kill Command (CLI)
+
+The `kill` command can send a signal termination (SIGTERM) command to any running processes. This process requires knowing the process IDs (PID) of the application(s) to be terminated. The `pidof` command will display the PID(s) associated with any running process.
+
+1. Press <kbd>Super</kbd> + <kbd>T</kbd> to launch a Terminal session.
+2. Type pidof + process or application name and hit Enter.
```
jacobgkau, line 45: Instead of having code blocks beneath each step, you could make any verbatim commands inline code with a single backtick (\`) on each side. If you're using code blocks so it's easier to copy/paste, though, that's perfectly fine.

content/ending-frozen-applications.md
```diff
@@ -0,0 +1,76 @@
...
+
```
jacobgkau, line 46: These code blocks need the excess empty lines removed similar to the earlier section.

content/ending-frozen-applications.md
```diff
@@ -0,0 +1,76 @@
...
+   ```bash
+   pidof firefox
```
jacobgkau, line 48: It can be tricky to guess what a process name is called and I could see value in explaining how to find the process name, but that's just a suggestion/not strictly necessary.

content/ending-frozen-applications.md
```diff
@@ -0,0 +1,76 @@
...
+   ```
+
+3. Take note of the listed PID(s).
+4. Type kill + PID(s) to kill all associated PID(s), then hit <kbd>Enter</kbd>.
+
+    ```bash
+    kill 20869 20862
+    ```
+
+Tip: Use the pipe <kbd>|</kbd> and `xargs` command to pipe multiple PIDs as arguments into the kill command:
```
jacobgkau, line 58: This is a handy trick. Is there any reason a user would need to list and copy/paste the PIDs as described above instead of just doing this in the first place?

content/ending-frozen-applications.md
```diff
@@ -0,0 +1,76 @@
...
+
+  ```bash
+  pidof firefox | xargs kill
+  ```
+
+## End All Running Processes
+
+In some instances even the Terminal or entire desktop environment can hang or freeze. Users can resolve this by entering `TTY` mode to enter commands. This method will use the `kill -9 -1` command, which kills all processes as allowed by the user account.
```
jacobgkau, line 66: `In some instances` is a dependent clause, so it should have a comma after it. (It would not need a comma if it was at the end of the sentence.)

I would add another `the` before `entire desktop environment` because it's not another name for `Terminal` (although this might be more of a stylistic choice.)

content/ending-frozen-applications.md
```diff
@@ -0,0 +1,76 @@
...
+
+1. Press <kbd>Ctrl</kbd> + <kbd>Alt</kbd> + <kbd>F5</kbd> to enter `TTY` mode.
+2. Enter the user name and password for the account.
+3. Type kill -9 -1 and hit <kbd>Enter</kbd>.
+
+    ```bash
+    kill -9 -1
```
jacobgkau, line 73: This is an interesting method. I personally use `sudo systemctl restart gdm3` in cases where this would be necessary, but this works too since `gdm3.service` has `Restart=always` set.

I wonder if it would be worth using `sudo` for this command; processes having to do with hardware (or anything else privileged) might need it.

jacobgkau (changes requested):
This article reads well overall. I've given feedback about style/grammar as well as the actual content.

> I haven't added the screenshot to /static/images yet, I'll need some help with that.

I saw you closed an earlier PR because of a typo, are you using the GitHub web interface instead of git? Let me know if you need assistance getting git/LFS set up or committing images.

content/ending-frozen-applications.md
jacobgkau: Actually, I didn't read this closely enough. You used the code block for the Firefox example.

Just add a single backtick on either side of `pidof` on this line since that specific word is a verbatim command.

content/ending-frozen-applications.md
Ampersandstorm: Good point, I'll apply suggested changes. I added the first part because there's always the "why would you do it _that_ way?" Linux guru, but on second glance don't think it really adds much to the article.

content/ending-frozen-applications.md
Ampersandstorm: I tried to get npm installed with Nathaniel but it kept giving some kind of error. Is there some way I can get this preview before making a pull request?

Ampersandstorm: I think I fixed the spacing, let me know if not

content/ending-frozen-applications.md
Ampersandstorm: according to kill's manpage it uses "TERM" (end) as the default, so this particular section would not be using the destructive -9. The section using TTY is sort of hail-mary for a completely frozen system. If -9 is bad practice I can add a suggestion to use the default (15). SIGSTOP I'm not familiar with, but appears to simply pause a running process. I'm not sure if that's useful for completely locked-up processes, I'll defer to your expertise.

content/ending-frozen-applications.md
Ampersandstorm: done, good catch. I also forgot the <kbd> tags for Enter

content/ending-frozen-applications.md
Ampersandstorm: I have implemented these changes - thanks!

content/ending-frozen-applications.md
Ampersandstorm: I believe I've fixed all instances of this

content/ending-frozen-applications.md
Ampersandstorm: yea good point

content/ending-frozen-applications.md
Ampersandstorm: thanks, good catch, I've updated the article to address this

content/ending-frozen-applications.md
Ampersandstorm: I tried to focus on "basic" commands since this article is geared toward new Linux users. I've added sudo to this command in the doc

content/ending-frozen-applications.md
Ampersandstorm: We can add it, I want to make sure I fixed the more critical items first

content/ending-frozen-applications.md
Ampersandstorm: image
removing those lines causes the linter to get angry at me, maybe I'm not understanding your comment

jacobgkau: That check in the linter is clearly not reasonable, since doing what it suggests changes the displayed formatting on the page. Let me see the best way to fix/work around that.

Ampersandstorm:
> This article reads well overall. I've given feedback about style/grammar as well as the actual content.
> 
> > I haven't added the screenshot to /static/images yet, I'll need some help with that.
> 
> I saw you closed an earlier PR because of a typo, are you using the GitHub web interface instead of git? Let me know if you need assistance getting git/LFS set up or committing images.

I'm using the github desktop app and vs code to pull/commit articles. I'm not sure how to go about adding folders and images to /static/images. Is it the same process?

jacobgkau:
> I'm using the github desktop app and vs code to pull/commit articles. I'm not sure how to go about adding folders and images to /static/images. Is it the same process?

I don't use either of those apps, but it sounds like they should both support LFS, based on [GitHub Desktop's documentation](https://docs.github.com/en/desktop/installing-and-configuring-github-desktop/configuring-and-customizing-github-desktop/about-git-large-file-storage-and-github-desktop) and some closed issues in VS Code's repository. Are you able to browse the existing images in the repo, or do they all show up as small text files? If you're able to browse existing images, then you should be able to add new ones. You'll just need to commit and push after adding them, which I assume those apps take care of for you.

If the existing images don't show up on your local system, then you should download Git LFS and run the install script: https://git-lfs.github.com/

content/ending-frozen-applications.md
jacobgkau: I disabled the MD031 check, the lint is now passing.

content/ending-frozen-applications.md
```diff
@@ -0,0 +1,81 @@
...
+
+Linux offers many methods to stop hanging or frozen processes. The System Monitor provides a GUI process management application with features similar to the Windows Task Manager, while the `kill` command can end single or multiple processes from the Terminal.
+
+>Caution: Suddenly terminating processes can result in unwanted data loss or system stability issues.
+
+## Close Specific Applications or Windows
+
+### Using System Monitor (GUI)
+
+The System Monitor comes prepackaged in Pop!\_OS. This application provides a GUI interface and functionality similar to the Windows Task Manager.
```
jacobgkau, line 30: I would still say `The System Monitor application`, `The System Monitor app`, or just `System Monitor`. I don't think having an article before this singular proper noun is correct grammar (but if you include a non-proper noun like `application` or `app`, now it's a definite article and correct.)

content/ending-frozen-applications.md
```diff
@@ -0,0 +1,81 @@
...
+2. Enter `gnome-system-monitor` and hit <kbd>Enter</kbd>.
+3. Locate frozen and hanging applications.
+   ![gnome-system-monitor](/images/ending-frozen-applications/gnome-system-monitor)
+4. Right click the application and click Stop in the dropdown menu.
+
+### Using the kill Command (CLI)
+
+The `kill` command provides several methods to terminate a process:
+
+- SIGTERM (15) - Sends the "end" command to a PID, this is the default method used if not number is specified
```
jacobgkau, line 42: `SIGTERM` would be a "terminate" signal, wouldn't it? I guess "end" is referring to what System Monitor calls it? This is under the CLI section, not the GUI section, so not sure how much it makes sense to go off of the GUI's label for a technical explanation like this.

Ampersandstorm:
> 



> > I'm using the github desktop app and vs code to pull/commit articles. I'm not sure how to go about adding folders and images to /static/images. Is it the same process?
> 
> I don't use either of those apps, but it sounds like they should both support LFS, based on [GitHub Desktop's documentation](https://docs.github.com/en/desktop/installing-and-configuring-github-desktop/configuring-and-customizing-github-desktop/about-git-large-file-storage-and-github-desktop) and some closed issues in VS Code's repository. Are you able to browse the existing images in the repo, or do they all show up as small text files? If you're able to browse existing images, then you should be able to add new ones. You'll just need to commit and push after adding them, which I assume those apps take care of for you.
> 
> If the existing images don't show up on your local system, then you should download Git LFS and run the install script: https://git-lfs.github.com/

I think I figured it out? I can see the images in my local repo. I created the folder and added the image locally (matching the path in the article), the github desktop app recognized it, I clicked some more buttons and I'm seeing my .png listed in the above changes.

content/ending-frozen-applications.md
```diff
@@ -0,0 +1,81 @@
...
+- SIGSTOP (19) - Sends the "stop" command to pause a process, processes are resumed with the SIGCONT (18)
+- SIGKILL (9) - Sends the "kill" command to a PID, this is the most destructive method, programs cannot ignore or block this command
```
jacobgkau, line 44: This list might look nicer if it used correct punctuation, only because the descriptions include multiple components and are long enough to be sentences anyway.

content/ending-frozen-applications.md
```diff
@@ -0,0 +1,81 @@
...
+## Close Specific Applications or Windows
+
+### Using System Monitor (GUI)
+
+The System Monitor comes prepackaged in Pop!\_OS. This application provides a GUI interface and functionality similar to the Windows Task Manager.
+
+1. Press <kbd>Super</kbd> + <kbd>T</kbd> to launch a Terminal session.
+2. Enter `gnome-system-monitor` and hit <kbd>Enter</kbd>.
+3. Locate frozen and hanging applications.
+   ![gnome-system-monitor](/images/ending-frozen-applications/gnome-system-monitor)
```
jacobgkau, line 35: This embed isn't working, the path needs the file extension (`.png`) at the end.

content/ending-frozen-applications.md
```diff
@@ -0,0 +1,81 @@
...
+
+The `kill` command provides several methods to terminate a process:
+
+- SIGTERM (15) - Sends the "end" command to a PID, this is the default method used if not number is specified
+- SIGSTOP (19) - Sends the "stop" command to pause a process, processes are resumed with the SIGCONT (18)
+- SIGKILL (9) - Sends the "kill" command to a PID, this is the most destructive method, programs cannot ignore or block this command
+
+`kill` takes arguments in the form of a Process ID (PID). The `pidof` command will display the PID(s) associated with any running process.
+
+1. Press <kbd>Super</kbd> and type "system monitor", hit enter to launch the program.
```
jacobgkau, line 48: It looks like this step got replaced in the wrong section.

I would add "then" after the comma and put Enter inside of `<kbd>` tags.

content/ending-frozen-applications.md
```diff
@@ -0,0 +1,81 @@
...
+2. Type `pidof` + process or application name and hit <kbd>Enter</kbd>.
+
+   ```bash
+   pidof firefox
+   ```
+
+3. Take note of the listed PID(s).
+4. Type kill + PID(s) to kill all associated PID(s), then hit <kbd>Enter</kbd>. This command uses the default SIGTERM (15) method. Multiple PIDs can be passed to `kill` by piping in PIDs using the `pidof` and `xargs` command:
```
jacobgkau, line 54: The steps to get the PID can probably be removed if the example doesn't use them. Just explain it so the user only needs to know the process name.

If you'd rather keep PIDs in play for some reason, then I'd change it back to how it was before-- the pipe method is not a progression from the PID steps and should not be in the same numbered list.

content/ending-frozen-applications.md
```diff
@@ -0,0 +1,73 @@
...
+
+Users can bypass a frozen desktop environment by entering `TTY` mode. This method will use the `kill -9 -1` command, which kills all processes as allowed by the user account. Users can alternatively use `kill -15 -1` which sends a less destructive SIGTERM (end) request to running processes.
+
+1. Press <kbd>Ctrl</kbd> + <kbd>Alt</kbd> + <kbd>F5</kbd> to enter `TTY` mode.
+2. Enter the user name and password for the account.
+3. Type the following command and hit <kbd>Enter</kbd>.
+    ```bash
+    sudo kill -15 -1
+    ```
+   >Note: Users can alternatively use `sudo kill -9 -1` which sends a more destructive SIGTERM (end) request to running processes.
```
jacobgkau, line 72: I don't think this is factually correct; `-9` uses SIGKILL.

I would recommend just keeping this how you had it before, using `-9` in the actual step, so this note isn't necessary.

jacobgkau (commented):
Sorry if you're still working on this, I started reviewing again when you requested re-review.

content/ending-frozen-applications.md
```diff
@@ -0,0 +1,68 @@
...
+facebookImage: /_social/article
+twitterImage: /_social/article
+
+hidden: false
+section:
+---
+
+## Closing Frozen Processes, Applications, and Desktops
+
+Linux offers many methods to stop hanging or frozen processes. The System Monitor provides a GUI process management application with features similar to the Windows Task Manager, while the `kill` command can end single or multiple processes from the Terminal.
```
jacobgkau, line 22: Same thing as before regarding "The System Monitor" (it was corrected in the section below, but it should probably be consistent here too.)

content/ending-frozen-applications.md
jacobgkau: Un-resolving this since it hasn't been addressed or defended; an earlier revision looked like it tried to switch to using the Launcher, but edited the wrong section.

content/ending-frozen-applications.md
```diff
@@ -0,0 +1,68 @@
...
+## Close Specific Applications or Windows
+
+### Using System Monitor (GUI)
+
+The System Monitor application comes prepackaged in Pop!\_OS. This application provides a GUI interface and functionality similar to the Windows Task Manager.
+
+1. Press <kbd>Super</kbd> + <kbd>T</kbd> to launch a Terminal session.
+2. Enter `gnome-system-monitor` and hit <kbd>Enter</kbd>.
+3. Locate frozen and hanging applications.
+   ![gnome-system-monitor](/images/ending-frozen-applications/gnome-system-monitor.png)
```
jacobgkau, line 35: Is pulseaudio supposed to be a hung application in this example? You could perhaps use something recognizable like Firefox to match the next section.

It might also be good if this screenshot includes the right-click menu that the next step refers to.

The framing for the screenshot looks a little strange-- not unacceptable, but inconsistent that there's a bit of space at the top but not on the sides. (And I don't know where the gradient at the top is coming from; might want to make sure the desktop wallpaper on your screenshot machine is a default one, if it's not already.)

content/ending-frozen-applications.md
```diff
@@ -0,0 +1,68 @@
...
+The `kill` command provides several methods to terminate a process:
+
+- SIGTERM (15) - Sends the "terminate" command to a PID (default)
+- SIGSTOP (19) - Sends the "stop" command to pause a process
+- SIGKILL (9) - Sends the "kill" command to a PID (destructive)
+
+`kill` takes arguments in the form of a Process ID (PID). The `pidof` command will display the PID(s) associated with any running process.
+
+1. Press <kbd>Super</kbd> + <kbd>T</kbd> to launch a terminal session.
+2. This command uses the default SIGTERM (15) method. Multiple PIDs can be passed to `kill` by piping in PIDs using the `pidof` and `xargs` command:
```
jacobgkau, line 49: This step is the only one that's written passively-- it's describing what can be done instead of telling the user what to do.

Suggested rephrase (feel free to do it differently):

> 2. Use the `pidof` command to get the PIDs of a process, and a pipe with the `xargs` command to input them into the `kill` command, which uses the default SIGTERM (15) method:

content/ending-frozen-applications.md
```diff
@@ -0,0 +1,68 @@
...
+    ```bash
+    pidof firefox | xargs kill
+    ```
+3. If the program remains unresponsive, use the SIGKILL (9) commend to force the program to close:
```
jacobgkau, line 53: You wrote `commend` where I think you meant `command`, although this is technically an option you're talking about, rather than an entire command.

content/ending-frozen-applications.md
jacobgkau: I'll resolve this one as out-of-scope for now, although I do think the usefulness of the rest of the instructions might be limited for some users without that knowledge.

`xprop | grep -i pid` will get the numeric PID of a GUI app, it looks like that would need to be cross-referenced with `ps` or something similar to get the name (if the name's necessary.)

jacobgkau: Actually, before we resolve this, on the topic of GUI apps where the user might not know the process name... did you consider `xkill` for inclusion in this article? It's much more point-and-click.

content/ending-frozen-applications.md
```diff
@@ -0,0 +1,77 @@
...
+## Close Specific Applications or Windows
+
+### Using System Monitor (GUI)
+
+The System Monitor application comes prepackaged in Pop!\_OS. This tool provides a GUI and functionality similar to the Windows Task Manager.
+
+1. Press <kbd>Super</kbd> and type `system monitor`, then hit Enter.
+2. Locate frozen and hanging applications.
+   ![gnome-system-monitor](/images/ending-frozen-applications/gnome-system-monitor.png)
+3. Right click the application and click Stop in the dropdown menu.
```
jacobgkau, line 35: Sorry, this just hit me now that it's included in the screenshot-- why are we stopping (pausing) the frozen process instead of ending or killing it?

content/ending-frozen-applications.md
```diff
@@ -0,0 +1,77 @@
...
+facebookImage: /_social/article
+twitterImage: /_social/article
+
+hidden: false
+section:
+---
+
+## Closing Frozen Processes, Applications, and Desktops
+
+Linux offers many methods to stop hanging or frozen processes. The System Monitor application provides a GUI with features similar to the Windows Task Manager, while the `kill` command can end single or multiple processes from the Terminal.
```
jacobgkau, line 22: The new `xkill` section looks great-- should it be included in the summary up here?

jacobgkau (commented):
This is looking great, just had a last question or two while reading it again.

jacobgkau (approved):
This looks great! Technically, `xkill` is a command similar to `kill`, but I think presenting it as a GUI option here works, since it uses the GUI point-and-click method to select the application to kill.

The article wasn't showing up on the main list because `section` wasn't set, so I added it to the Software Troubleshooting section.

PR #916, opened 2022-02-10
Add kudu6 tech docs to list

jacobgkau:


ahoneybun (commented):
Everything works and looks good when testing it live.

ahoneybun (approved):
Everything looks good live!

PR #918, opened 2022-02-14
Find Help Using the Terminal article and images

Ampersandstorm:
Created an article describing several methods to find information about commands, this is an entry-level topic that can referenced by similar content.


content/finding-command-help.md
```diff
@@ -0,0 +1,97 @@
...
+```
+![man output](/images/finding-command-help/man-output.png)
+
+### info
+
+`info` pages are similar to man pages, however their descriptions for command usage are more in-depth and include examples of output. Users can also navigate to other related info pages using hypertext links within the info page. `info` opens an info page in the Terminal session, pressing <kbd>q</kbd> exits the manpage and returns the user to the shell prompt.
+
+**Use case**: A user wants to see examples of command usage, command output, and easily navigate to commands offering related or supporting functionality.
+
+**Usage**: Type `info` + any command into a Terminal session, then hit Enter. This example uses the `ls` command.
```
jacobgkau, line 57: Since you cited hypertext links as a feature of `info`, you may want to explain how to use the arrow/tab keys and enter/spacebar keys to navigate the links. Users would probably default to trying to click, which will not work.

content/finding-command-help.md
```diff
@@ -0,0 +1,97 @@
...
+
+`apropos` performs a keyword search on command man pages. This is useful when searching for a command by its function. `apropos` prints output to the terminal session without navigating away from the shell prompt.
+
+**Use case**: A user can't remember the specific name of a command, but knows keywords commonly used to describe the command's functions.
+
+**Usage**: Type `apropos` + keyword. This example searches for commands related to the `delete` keyword.
+```bash
+apropos delete
+```
+![apropos output](/images/finding-command-help/apropos-output.png)
```
jacobgkau, line 73: Not sure this example demonstrates good value in using `apropos` since you searched for "delete" and the actual command to delete files (`rm`) wasn't included in the search results. `unlink` usually behaves similarly to `rm`, but its description ("possibly the file") doesn't really make that clear.

content/finding-command-help.md
```diff
@@ -0,0 +1,97 @@
...
+
+### --help
```
jacobgkau, line 75: This would be open for discussion, but I'd personally think `--help` would be more consistently useful than any of the options above it, and for that reason should be listed higher up on the page.

content/finding-command-help.md
```diff
@@ -0,0 +1,97 @@
...
+
+**Use case**: A user wants to quickly reference all options for a known command in the current Terminal session.
+
+**Usage**: Type a command + `--help`. This example uses the `cd` command. Commands must be an exact match.
+```bash
+cd --help
+```
+![--help output](/images/finding-command-help/--help-output.png)
+
+### whatis
```
jacobgkau, line 87: Wondering if `which` + `ls -l` + `dpkg -S` to find out which package is providing a command would be useful to document here.

Along that line, information or tips for finding command info online may be helpful. (https://manpages.ubuntu.com/ or https://manpages.debian.org/ can be nicer to browse than doing everything in the terminal-- it's good to know the emergency options for when you're stuck in a TTY and have no internet or other device to read from, but that's a less common situation.)

content/finding-command-help.md
```diff
@@ -0,0 +1,97 @@
...
+```
+![info output](/images/finding-command-help/info-output.png)
+
+### apropos
+
+`apropos` performs a keyword search on command man pages. This is useful when searching for a command by its function. `apropos` prints output to the terminal session without navigating away from the shell prompt.
+
+**Use case**: A user can't remember the specific name of a command, but knows keywords commonly used to describe the command's functions.
+
+**Usage**: Type `apropos` + keyword. This example searches for commands related to the `delete` keyword.
```
jacobgkau, line 69: You included "then hit Enter" in the first three sections, but not the last three. I think it should be consistent one way or the other. (If you go with it, consider putting "Enter" in `<kbd>` tags.)

content/finding-command-help.md
```diff
@@ -0,0 +1,97 @@
...
+
+**Usage**: Type `man` + any command into a Terminal session, then hit Enter. This example uses the `reboot` command.
+```bash
+man reboot
+```
+![man output](/images/finding-command-help/man-output.png)
+
+### info
+
+`info` pages are similar to man pages, however their descriptions for command usage are more in-depth and include examples of output. Users can also navigate to other related info pages using hypertext links within the info page. `info` opens an info page in the Terminal session, pressing <kbd>q</kbd> exits the manpage and returns the user to the shell prompt.
```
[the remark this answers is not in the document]
Ampersandstorm: Regarding the 1st sentence: I believe a comma would be correct since this is joining an independent clause using a conjunction. I typically reference Strunk & White's "Elements of Style" (4th edition), which is generally considered a good source of truth. I think rule #4 applies here: http://www.jlakes.org/ch/web/The-elements-of-style.pdf. Let me know if you find that acceptable.

content/finding-command-help.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Find Command Help in the Linux Terminal
+
+Even the most experienced Linux users have difficulty remembering proper syntax for Terminal commands and command options. All Linux shells support common utilities that search for and explain commands for a given task. These commands may navigate the user to a text file or navigable structure of help files, while others simply print output into the current Terminal session.
+
+The commands below are entered in a Terminal session. In Pop!\_OS, Terminal sessions can be started by pressing <kbd>Super</kbd> + <kbd>t</kbd>.
+
+### help
+
+The `help` command displays all built-in commands supported by the current shell.
```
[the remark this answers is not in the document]
Ampersandstorm: removed

content/finding-command-help.md
```diff
@@ -0,0 +1,97 @@
...
+
+facebookImage: /_social/article
+twitterImage: /_social/article
+
+hidden: false
+section: about-your-os
+tableOfContents: true
+---
+
+## Find Command Help in the Linux Terminal
```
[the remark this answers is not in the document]
Ampersandstorm: removed

content/finding-command-help.md
```diff
@@ -0,0 +1,97 @@
...
+
+**Usage**: Type `help` into a Terminal session, then hit Enter.
+```bash
+help
+```
+![help output](/images/finding-command-help/help-output.png)
+
+### man
+
+`man` is short for manual pages. A command's manual page is a consistently structured text file describing the command's function, usage syntax, options, and similar commands. `man` opens a manpage in the terminal session, pressing <kbd>q</kbd> exits the manpage and returns the user to the shell prompt.
```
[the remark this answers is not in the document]
Ampersandstorm: updated

content/finding-command-help.md
Ampersandstorm: added a table with the basic navigation keys

content/finding-command-help.md
Ampersandstorm: Man and Info are the resources that apropos, --help, and whatis pull from, so I thought it made sense to present them in that order. I agree the article would have more functional value if --help was closer to the top. I've reorganized the content so that the less in-depth options are presented first.

content/finding-command-help.md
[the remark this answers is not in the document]
Ampersandstorm: fixed

content/finding-command-help.md
jacobgkau: You're correct about apropos and whatis pulling from man pages, but `--help` does not pull from man pages. It's a flag and is handled by whatever application you're running it on. Most of our own utilities support `--help` but do not have man or info pages (see pop-upgrade or kernelstub for examples of this.)

content/finding-command-help.md
Ampersandstorm: switching out "delete" for "reboot".

content/finding-command-help.md
Ampersandstorm: added tags and mentioned hitting Enter for all sections

content/finding-command-help.md
```diff
@@ -0,0 +1,97 @@
...
+twitterImage: /_social/article
+
+hidden: false
+section: about-your-os
+tableOfContents: true
+---
+
+## Find Command Help in the Linux Terminal
+
+Even the most experienced Linux users have difficulty remembering proper syntax for Terminal commands and command options. All Linux shells support common utilities that search for and explain commands for a given task. These commands may navigate the user to a text file or navigable structure of help files, while others simply print output into the current Terminal session.
```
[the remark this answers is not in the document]
Ampersandstorm: changed wording

content/finding-command-help.md
Ampersandstorm: I didn't consider "which", but it is a good addition and will help set this article apart from the dozens of similar "finding help" resources that pop up in a Google search. I ran the command you mentioned "which ls -l dpkg -S" and noticed the output is the same as just "which ls dpkg". Is that expected?

jacobgkau: They were three separate commands. The use case goes like this: you know a command, and you want to know which package the command comes from (so you can install/uninstall it, look up documentation, etc.) First step, get path of the command:

```
jacob@serw12:~$ which libreoffice 
/usr/bin/libreoffice
```

Second, put that path into dpkg to search for what package provides it:

```
jacob@serw12:~$ dpkg -S /usr/bin/libreoffice 
libreoffice-common: /usr/bin/libreoffice
```

The issue with that is, dpkg only knows where the package wanted to install itself to, so if it's been symlinked somewhere else on your system, you may need to see where it's linking from and search for the original path instead of the one you're using:

```
jacob@serw12:~$ which shutdown
/usr/sbin/shutdown
jacob@serw12:~$ dpkg -S /usr/sbin/shutdown
dpkg-query: no path found matching pattern /usr/sbin/shutdown
jacob@serw12:~$ ls -alh /usr/sbin/shutdown
lrwxrwxrwx 1 root root 14 Jan 19 12:17 /usr/sbin/shutdown -> /bin/systemctl
jacob@serw12:~$ dpkg -S /bin/systemctl
systemd: /bin/systemctl
```

Unfortunately, even that sometimes doesn't work, because if a package requests to install itself into Dir A and your system has Dir A symlinked to Dir B, `which` and `ls` will both return Dir B, but `dpkg` will only know about Dir A. In those cases, you just have to know the common symlinks, like /bin to /usr/bin, /lib to /usr/lib, etc.

```
jacob@serw12:~$ which mkdir
/usr/bin/mkdir
jacob@serw12:~$ dpkg -S /usr/bin/mkdir
dpkg-query: no path found matching pattern /usr/bin/mkdir
jacob@serw12:~$ ls -alh /usr/bin/mkdir
-rwxr-xr-x 1 root root 91K Dec  5  2020 /usr/bin/mkdir
jacob@serw12:~$ ls -alh /bin
lrwxrwxrwx 1 root root 7 Oct 22  2020 /bin -> usr/bin
jacob@serw12:~$ dpkg -S /bin/mkdir
coreutils: /bin/mkdir
```

jacobgkau: I don't know how much of this needs to be included in this article, I just know looking up the package that owns a command is something I've done a number of times (on Pop systems with `dpkg -S` and with other package managers on other systems), and that's probably my top use case for `which` as well.

content/finding-command-help.md
```diff
@@ -0,0 +1,109 @@
...
+
+### which
+
+`which` provides the directory location of a command.
+
+**Use case**: A user would like to know which installed package provides a given command.
+
+**Usage**: Type `which` + command(s), then hit <kbd>Enter</kbd>. This example uses the ls and dpkg commands.
+```bash
+which ls -l dpkg -S
```
jacobgkau, line 59: Including flags in a `which` command doesn't make sense; see my last two comments on the conversation thread and feel free to modify/remove/update however you see fit.

content/finding-command-help.md
Ampersandstorm: Thanks for clarifying. Definitely useful, although I need to familiarize myself with this method. I'll add some version of this and make note of the symlink considerations.

content/finding-command-help.md
```diff
@@ -0,0 +1,109 @@
...
+![apropos output](/images/finding-command-help/apropos-output.png)
+
+### man
+
+`man` is short for manual pages. A command's manual page is a consistently structured text file describing the command's function, usage syntax, options, and similar commands. `man` opens a manpage in the terminal session, and pressing <kbd>q</kbd> exits the manpage and returns the user to the shell prompt.
+
+**Use case**: A user wants to see actions performed by a specific command, and available options to modify the outcome of the command.
+
+**Usage**: Type `man` + any command into a Terminal session, then hit <kbd>Enter</kbd>. This example uses the `reboot` command.
+```bash
```
storyaddict, line 82: There need to be line-breaks between **Usage** line, codeblock and png, otherwise it may not pass linter.

jacobgkau: It has already passed the linter; we had to [disable that lint](https://github.com/system76/docs/pull/915#discussion_r802975777) because it prevented having code blocks within ordered or unordered list items.

However, it would still be good practice to have the line break in cases like this where it's not part of a list item.

Ampersandstorm: Thanks for clarifying

content/finding-command-help.md
Ampersandstorm: While poking around with this I noticed `which libreoffice | xargs dpkg -S` returns the correct package name, however `which libreoffice | xargs ls -alh` shows the path we passed to `dpkg -S` in the prior command (/usr/bin/libreoffice) is actually symlinked to ../lib/libreoffice/program/soffice. Is the concern that the symlinked files will return the "no matching path found" error, the user will assume the incorrect path, or both? I took my best guess in the most recent draft.

jacobgkau: What's going on with `libreoffice` is the following:
- dpkg installs the actual executable, `soffice`, to `/usr/lib/libreoffice/program/soffice`.
- dpkg also installs a symlink to `/usr/bin/libreoffice`; whoever created the LibreOffice packaging knew that people would want to start it with the command `libreoffice`, so they included that symlink in the package.
- When you ask dpkg (using the `-S` search flag) what package owns the file `/usr/bin/libreoffice`, you're asking it what package owns the symlink; because that symlink came from the package, dpkg is able to return the package name, even though it's not the actual path of the executable.

The concern would be the `no path found matching pattern ...` error. I think your last sentence "This verification step does not account for additional potential scenarios where a package installs to symlinked directory." is acceptable (if you don't want to go into the details of [the `/usr` merge](https://www.freedesktop.org/wiki/Software/systemd/TheCaseForTheUsrMerge/).) However, I would choose a different example for the `xargs ls -alh` command (pick one that _doesn't_ work with `dpkg -S` but does work after following the `ls` trail, such as `shutdown` linking to systemd from my comment above).

Also, your last paragraph refers to an `openoffice` command where you were trying to say `libreoffice`.

Ampersandstorm: Thanks for clarifying. This process provides good info but introduces a lot of concepts (path environment variables, /user merge, symlinks, etc.). In the latest draft I've tried to include it without going too far down the rabbit hole.

content/finding-command-help.md
```diff
@@ -0,0 +1,131 @@
...
+whatis mkdir
+```
+![whatis output](/images/finding-command-help/what-output.png)
+
+### which & dpkg -S
+
+`which` provides the directory location of a command. The output of `which` can be passed into `dpkg -S` to perform a search of associated packages.
+
+**Use case**: A user is attempting to learn about commands included with a recently installed program, and would like to know which installed package provides a specific command
```
jacobgkau, line 57: This sentence is missing punctuation at the end.

You're stating two different use cases here, at least how I'm reading it. The first one ("attempting to learn about commands included with a recently installed program") is easy to do with `dpkg -S <packagename>`, but it's not covered in this section. You're really only covering the second one.

content/finding-command-help.md
```diff
@@ -0,0 +1,131 @@
...
+```
+![whatis output](/images/finding-command-help/what-output.png)
+
+### which & dpkg -S
+
+`which` provides the directory location of a command. The output of `which` can be passed into `dpkg -S` to perform a search of associated packages.
+
+**Use case**: A user is attempting to learn about commands included with a recently installed program, and would like to know which installed package provides a specific command  
+
+**Usage**: Type `which` + command, then pipe the output of `which` into `dpkg -S` using `xargs`. The `xargs` command allows users to pass the output of a command as standard input into another command. The below example searches for the path of the `libreoffice` command, and then searches for the program associated with that path.
```
jacobgkau, line 59: The last sentence should say `package associated with that path` instead of `program associated with that path`. A package can contain more than one program (or a complex program can be split between multiple packages); the installed executable file is the program.

content/finding-command-help.md
```diff
@@ -0,0 +1,131 @@
...
+
+```bash
+which libreoffice | xargs dpkg -S
+```
+![which output](/images/finding-command-help/which-output.png)
+
+Users may see an error stating no matching path can be found. This occurs when `which` returns a symbolic link instead of the actual installation directory. Packages create symbolic links because the Terminal only recognizes executable commands from specific default and user defined directories. `dpkg` only knows the intended default installation path for a package, but may not know which package owns a symbolic link. Pass the path output from `which` into the `ls -alh` command to verify if an executable file is symbolically linked.
```
jacobgkau, line 66: > This occurs when `which` returns a symbolic link instead of the actual installation directory. Packages create symbolic links because the Terminal only recognizes executable commands from specific default and user defined directories.

- `which` returning a symbolic link is not the cause of this; `which` returning a symbolic link (or any file) that wasn't included in a package is the cause of this.

> Packages create symbolic links because the Terminal only recognizes executable commands from specific default and user defined directories.

This is true, but not entirely relevant to the error message at hand for the above reason.

> `dpkg` only knows the intended default installation path for a package, but may not know which package owns a symbolic link.

This is a misrepresentation of the issue. Again, if a package owns a symbolic link, `dpkg` will know which one it is. The issue is if a package installs into a directory (such as `/bin`) that had already been symlinked to another directory (`/usr/bin`) before installation.

This misunderstanding might be partially my fault for the examples I provided. For the shutdown example, I was actually unaware _what_ was creating the `/usr/sbin/shutdown` file-- I assumed it was being created by a post-install script or systemd itself, rather than being packaged. However, I did some digging and discovered that `/usr/sbin/shutdown` is actually installed to `/sbin/shutdown`, which dpkg recognizes as installed from the `systemd-sysv` package. So it's actually the same issue as the one I showed with `mkdir` earlier.

content/finding-command-help.md
```diff
@@ -0,0 +1,131 @@
...
+
+```bash
+which shutdown | xargs ls -alh
+```
+![which ls](/images/finding-command-help/which-ls.png)
+
+The output indicates the package containing the `open` command is symbolically linked to `/bin/systemctl`. Now pass `/bin/systemctl` into the `dpkg -S` command to determine the package that owns this directory.
```
jacobgkau, line 73: You referred to an `open` command that is not referenced anywhere else in the article.

jacobgkau (commented):
See comments. I'm currently writing up a suggested change and will commit it, then you can pare it down as you see fit.

jacobgkau:
@Ampersandstorm Feel free to pare down what I committed as you see fit, but this is more accurate and likely more useful. I tried to follow the style you used for the other sections.

Ampersandstorm:
@jacobgkau Thanks for fixing that section. I think this topic could be its own article (maybe as a step to troubleshooting installed packages), but does serve a purpose in this context. We can always split it off if we want a more fleshed out version that really goes into detail about package management and symlinking.

jacobgkau (approved):

PR #920, opened 2022-02-17
fixed important typos in code snippets

Fibs7000:
I upgraded from 20.10 lately and these two typos had made it quite a bit harder. (especially the first one)


jacobgkau (approved):
Thank you for correcting these typos.

PR #921, opened 2022-02-19
Specifying the whole DaVinci File name makeresolve

jesaspiring:
Using Lite or Studio doesn't work anymore at least for me. What I did try was running this command instead and it worked. ./makeresolvedeb*.sh DaVinci_Resolve_17.4.3_Linux.run


jacobgkau:
The syntax has indeed changed. I'd like to keep the instructions copy/paste-able wherever possible. Since this command is being run in a directory created during the tutorial, we can use a glob to select the .run file similar to the one already used to select the .sh file.

jacobgkau (approved):
The section now looks like this:

image

Confirmed the new command works on a test machine.

PR #923, opened 2022-02-21
Add CPU thermal information

garnerlogan65:
Add introductory sentence, and thermal testing text from unexpected-reboots.md


storyaddict (changes requested):
Heading for "Drive Issues" was accidentally grabbed from Unexpected Reboots article. Please remove that, and also adjust headers so that we don't go from ## to #### per lint checks, then this can be merged.

storyaddict (approved):
Removed accidentally copied over header.

PR #924, opened 2022-02-23
route typo fix

LobbyLobster:
My system broke after latest Nvidia driver updates and while trying to fix the problem realized package manager repair commands link was forwarding to the wrong route.

### DIFF
--- a/content/pop-recovery.md
+++ b/content/pop-recovery.md
@@ -119,7 +119,7 @@ sudo cp -n /etc/resolv.conf /mnt/etc/
 sudo chroot /mnt
 ```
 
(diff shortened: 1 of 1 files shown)


LobbyLobster:
Apparently I also had to add .md extension and this is not the only wrong route all links forwarding to /article/ are broken since it doesn't exists.

ahoneybun:
The missing / was the correct fix as you can see the other links work with it. Thank you for this PR and letting us know about the issue!

ddotthomas:
I just closed a pull request I made as I hadn't checked for this one. I think changing article/package-manager-pop to /articles/package-manager-pop will be the easiest fix.

ahoneybun:
@LobbyLobster we'll want to keep this open as we can use it to fix this issue.

ahoneybun (approved):
LGTM :tada:

LobbyLobster:
Sorry I'm not accustomed to contributing. Thought it's already done and I had to close it. Thanks for letting me know!

PR #926, opened 2022-02-23
Update fixing-drives.md

garnerlogan65:
make it clear that fsck must be run on a partition, not the entire drive.


ahoneybun (approved):
LGTM!

PR #927, opened 2022-02-26
Fix broken "additional setup" link.

Kennegh:
The "additional setup" href at the end of both the shared drive and dedicated drive sections is missing the "-not-in-vm" portion of the ID for the relevant section (line 469).

### DIFF
--- a/content/windows.md
+++ b/content/windows.md
@@ -339,7 +339,7 @@ After completing the wizard, Windows will display a slideshow while it finishes 
 
 ![Windows 10 desktop](/images/dual-booting/windows-10-desktop.jpg)
 
-See [additional setup](#additional-setup-for-windows) for next steps.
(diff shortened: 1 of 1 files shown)


ahoneybun:
It looks like this was fixed already since the links work on the live site.

Kennegh:
Neither of the "See additional setup for next steps." links at the end of
https://support.system76.com/articles/windows/#performing-the-installation
or
https://support.system76.com/articles/windows/#performing-the-windows-installation
work for me. Both are pointing to 
https://support.system76.com/articles/windows/#additional-setup-for-windows
which does not exist on the page. It should be pointing to
https://support.system76.com/articles/windows/#additional-setup-for-windows-not-in-vm
which is what the Table of Contents has for the additional setup section.

PR #930, opened 2022-03-03
fix: Typo in transition-firmware.md

sudomakecoffee:
Small editing change to correct a typo in description of what is and is not functionally possible in S76 Open Firmware


jacobgkau (approved):
Thank you for the typo fix.

PR #945, opened 2022-03-09
Updates for 21.10

thomas-zimmerman:
Ran through commands in 21.10 for hibernation to verify that they continue to work

* expend on security benefit for hibernation over suspend: thanks to @ChristopherRogers1991
* update for clarity on point where you need to reboot
* add  #comments for some of the commands where you edit config files in bash


storyaddict (approved):
Made what would have been requested changes with my current commits. Ready to merge now.

PR #947, opened 2022-03-09
Update add-a-printer-ubuntu.md

Seibz:
added brother printer instructions to ubuntu article

### DIFF
--- a/content/add-a-printer-ubuntu.md
+++ b/content/add-a-printer-ubuntu.md
@@ -55,7 +55,7 @@ And when adding the printer, select type of connection as **DNS-SD**, and then *
 
 Brother provides a driver installation tool for Linux users. Install the appropriate driver for your Brother printer by downloading this tool and running the installer with your printer model appended to the command.
 
-1. Click [this link](https://support.brother.com/g/b/productsearch.aspx?c=us&lang=en&content=dl) to download Brother's Driver Install Tool for Linux while searching for your appropriate printer.
+1. Click [this link](https://support.brother.com/g/b/productsearch.aspx?c=us&lang=en&content=dl) to download Brother's Driver Install Tool for Linux while searching for the appropriate printer.
 
 2. Click the "Click here to download the tool" link on the Brother download page.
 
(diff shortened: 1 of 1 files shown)


content/add-a-printer-ubuntu.md
```diff
@@ -51,6 +51,42 @@ sudo usermod -aG lpadmin $USER
 
 And when adding the printer, select type of connection as **DNS-SD**, and then **Forward**, and then **ESC/P-R**.
 
+## Brother Printers
+
+Brother provides a driver installation tool for Linux users. Install the appropriate driver for your Brother printer by downloading this tool and running the installer with your printer model appended to the command.
+
+1. Click [this link](https://support.brother.com/g/b/productsearch.aspx?c=us&lang=en&content=dl) to download Brother's Driver Install Tool for Linux while searching for your appropriate printer.
```
ahoneybun, line 58: I believe this should be "for the appropriate printer"

content/add-a-printer-ubuntu.md
```diff
@@ -51,6 +51,42 @@ sudo usermod -aG lpadmin $USER
...
+
+2. Click the "Click here to download the tool" link on the Brother download page.
+
+3. Select Linux (deb) for your "OS Version", then click OK.
+
+4. Click "Agree to the EULA and Download".
+
+5. Choose the Save File option when prompted.
+
+6. Press <kbd>Super</kbd> + <kbd>T</kbd> to launch the Terminal application.
```
ahoneybun, line 68: This is the shortcut for Pop not Ubuntu so we need to change that.

ahoneybun (changes requested):

content/add-a-printer-ubuntu.md
storyaddict: I think either works, but I think "the appropriate" sounds better.

Seibz: I would agree

content/add-a-printer-ubuntu.md
Seibz: We would, do we have a different super key image?

Seibz (commented):
Looks good

content/add-a-printer-ubuntu.md
storyaddict: I've just been using Super or SUPER like we have here. That way it's agnostic.

storyaddict (approved):
Fixed wording issue.

ahoneybun (approved):

PR #948, opened 2022-03-09
Update linux-gaming.md

Seibz:
added a section detailing new functionality for the protonup tool

### DIFF
--- a/content/linux-gaming.md
+++ b/content/linux-gaming.md
@@ -99,7 +99,7 @@ Downloaded 100.0% - 370.81 MiB/370.81 MiB
 [INFO] Installed in: /home/aaronh/.steam/root/compatibilitytools.d/Proton-6.10-GE-1
 ```
 
-You can also use `protonup` to install older versions of ProtonGE if you have a game that works better on an older release by using the `protonup --releases` command. The `protonup -t` command can then be run, followed by a specific release, to grab that release:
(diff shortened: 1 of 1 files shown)


content/linux-gaming.md
```diff
@@ -99,6 +99,18 @@ Downloaded 100.0% - 370.81 MiB/370.81 MiB
 [INFO] Installed in: /home/aaronh/.steam/root/compatibilitytools.d/Proton-6.10-GE-1
 ```
 
+You can also use `protonup` to install older versions of ProtonGE if you have a game that works better on an older release by using the `protonup --releases` command and then the `protonup -t` command followed by a specific release, to grab that release:
```
storyaddict, line 102: This is a run-on sentence.

"...install older versions of ProtonGE. For example, if you have a game that worked better on a previous version, you can specify the version to install by using the `protonup --releases` command." Etc.

when in doubt, break it down. Also, we should mention that protonup-qt is available in the Pop shop as a GUI alternative.

storyaddict (changes requested):
See in-line comment.

content/linux-gaming.md
storyaddict: The first sentence is still a run-on.

storyaddict (approved):
Nice addition.

PR #955, opened 2022-03-15
Moved the Disclaimer section back to the top

gangwerz:
Disclaimer was bumped down to the 3rd section, which kind of defeats the purpose of a disclaimer warning about the scope of support.


garnerlogan65 (approved):
Perfect. Good to have scope at the top.

PR #956, opened 2022-03-17
Update install-in-vm.md

hl2guide:
Mentions the required EFI mode needed for a successful install and boot.


jacobgkau (approved):

jacobgkau:
It's worth noting that it looks like this article will be removed entirely as part of https://github.com/system76/docs/pull/958. In the meantime, I'm merging this improvement.

PR #959, opened 2022-03-17
collapse formating

storyaddict:
This PR Does the following:

- Adds collapsable section HTML to a few articles.

- Subsequent changes will be done on a per article basis per Thomas' request.


storyaddict:
Can we get a change made @leviport or @ahoneybun to allow for these in-line HTML rules so we can pass lint?

storyaddict:
It looks like it only sent one of the changed docs (with the collapse-able sections...odd. Well that is more in-line with what Thomas wanted anyway.

jacobgkau (changes requested):
I'm not sure if we need to get UX involved in this, but it seems kind of weird to have all of the content hidden with only headings visible by default.

Screenshot from 2022-03-18 13-42-02

I end up having to click a lot just to read the article. The nested expanders in e.g. the Deja Dup section is certainly too much.

https://user-images.githubusercontent.com/7199422/159073563-a046920d-eb52-41e3-b501-d282d59bda1f.mp4

This also breaks the Table of Contents; clicking on a subsection doesn't work if the subsection hasn't been expanded yet.

gangwerz:
To make the sections open by default, you can use the below form of the `<details>` tag with the `open` attribute added.
```html
<details open>
....
</details>
```

storyaddict:
So after consulting with @thomas-zimmerman and @gangwerz , we still want the collapse-able sections as an option. Not for every segment though. We *do* want the ability to navigate to sections from the ToC even if collapsed, if at all possible.

Ampersandstorm (commented):
Not sure if my review is needed here, but I'm on board with Jacob's suggestions regarding this feature.

storyaddict:
I'm fine with "open" being the rule not the exception, but we do want some particular sections of certain articles to start collapsed. Going to play with this now.

gangwerz:
PR #972 may change the trajectory of these changes.

storyaddict:
Closing these collapsible section PRs for now as we may be moving to mdbook.

PR #960, opened 2022-03-18
Audio collapsable

storyaddict:
This PR adds expandable sections to a few documents (which were supposedly in the previous PR), with a focus on the audio article.

### DIFF
--- a/content/add-a-printer-pop.md
+++ b/content/add-a-printer-pop.md
@@ -46,7 +46,6 @@ sudo usermod -aG lpadmin $USER
 
 </details>
 
-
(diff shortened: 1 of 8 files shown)


garnerlogan65 (commented):
Given the length of the text, I think we may need to consider how we want to handle sections. I don't know if several sections are friendly from a user perspective if the overall text is short. Being collapsable by default may hide otherwise useful information depending on the issue that is experienced.

storyaddict:
Closing these collapsible section PRs for now as we may be moving to mdbook.

PR #964, opened 2022-03-18
include how to restore img firmware

garnerlogan65:
add img instructions

### DIFF
--- a/content/flash-firmware-from-usb.md
+++ b/content/flash-firmware-from-usb.md
@@ -46,6 +46,18 @@ The highlighted files pictured below (the hash will change based on file and mod
 
 ![files-unzipped](/images/open-firmware-smmstore/files-unzipped.png)
 
+## Extracting an img to the USB drive
+
+Firmware is typically packaged in the form of a zip archive, but support will ocasionally provide an .img image. If the firmware image provided is a .img file, it will need to be flashed using the Disks application. We do this by "restoring" the image to the drive.
+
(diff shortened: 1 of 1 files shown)


content/flash-firmware-from-usb.md
```diff
@@ -46,6 +46,18 @@ The highlighted files pictured below (the hash will change based on file and mod
 
 ![files-unzipped](/images/open-firmware-smmstore/files-unzipped.png)
 
+## Extracting an img to the USB drive
+
+Firmware is typically packaged in the form of a zip archive, but support will ocasionally provide an .img image. If the firmware image provided is a .img file, it will need to be flashed using the Disks application. We do this by "restoring" the image to the drive.
```
leviport, line 51: Think it's worth mentioning Popsicle? It's quite a bit more user-friendly than using Disks.

storyaddict: I think we should include Popsicle. Yes.

garnerlogan65: To be honest, I didn't consider that. I pulled most of the text from the live disk article, which we'll also be re-working. I agree with this change. I'll work on this.

ahoneybun:
Updated the article with Pop and Ubuntu steps (Pop has Popsicle and Ubuntu has Disks). Also update images for Popsicle (and the desktop) as we are sharing the images in the pop-live-disk article.

thomas-zimmerman (approved):
The changes look good to me in running `npm start`

PR #970, opened 2022-03-21
Added note about Windows games on NTFS drives for Steam/Proton

ottworks:


Ampersandstorm (changes requested):
Removed an exclamation point in the linking syntax to resolve some strange formatting and incorrectly rendered link.
image

gangwerz (approved):
Changes look good, and will add clarity for the foot notes. :+1:

PR #974, opened 2022-03-24
Lowering File Sizes of GIF Assets

gangwerz:
I am looking to minimize the size of the docs repo. The GIFs were recorded at 720p, 30FPS. 

Framerate seemed the best way to reduce size while maintaining legibility. I tried lowering to 480p, but the text in the result was not remotely readable.

**Savings**
Chrome
- Original: 37M
- 7 FPS: 9M
Firefox
- Original: 83M
- 7 FPS: 21M


gangwerz:
For posterity, I used FFMPEG to do the converting with the command below.

```bash
ffmpeg -i <filename>.gif -vf "fps=7" <filename>-7.gif
```

gangwerz:
I was originally going to try to compress Images further as well, but we can address that as needed on a per-doc basis.

The largest files, the 2 GIFs, have been shrunk a good amount with the above `ffmpeg` command.

storyaddict (approved):
Changes look good. Thanks @gangwerz

jacobgkau (approved):
Seeing these two gifs used in the article without accompanying text raises questions about the usefulness of text articles vs. video tutorials. If I had written this article, I would have had numbered or bullet-pointed lists for the buttons to click instead of having gifs at all, but I can understand that it may be easier for someone to just see where to click and follow along.

Anyway, the framerate reduction looks good. The cursor is still visible to see what's being clicked on.

Issue #975, opened 2022-03-26
format-drive live disk details


garnerlogan65:
The article mentions that a live-disk may need to be used, but it doesn't provide a link to the live disk article or much of a description for novice users.

PR #977, opened 2022-03-29
added subheaders for each option and linked to appropriate article

Ampersandstorm:
Updated per feedback in #975 - added sections to explain each option and link to the appropriate articles.

### DIFF
--- a/content/format-drive.md
+++ b/content/format-drive.md
@@ -26,7 +26,13 @@ tableOfContents: true
 
 In order to format the drive that has your OS on it we will need to boot from either a live disk or the Pop Recovery partition.
 
-**NOTE** Formatting the entire OS drive (including the Recovery partition) requires booting from a live disk or another drive.
+### Option 1: Live USB
+
+A live disk is a bootable USB storage device that runs the entire operating system in RAM. See [this article for instructions to create a bootable Pop!\_OS live USB.](/articles/live-disk/)
+
+### Option 2: Pop Recovery Partition
+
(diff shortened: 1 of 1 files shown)


content/format-drive.md
```diff
@@ -36,6 +42,6 @@ In this screenshot the drive is a USB Flash Drive, though the concept is the sam
 
 ![Disks](/images/format-drive/disks-format-disk.png)
 
-Now select the **Format Disk...** option then make sure to have the 'Compatible with modern systems and hard drives > 2TB (GPT)' selected (you can use either the Quick option or the Slow option).
+Now select the `Format Disk` option then make sure to have the 'Compatible with modern systems and hard drives > 2TB (GPT)' selected (you can use either the Quick option or the Slow option).
```
ahoneybun, line 45: Is there a reason to change the style here? Do we have an updated style somewhere?

ahoneybun (changes requested):

content/format-drive.md
jacobgkau: This is correct use of the Markdown inline code element. Using bold for word-for-word references to anything is not a standard. I agree with this change.

That said, I would still include the ellipses within the inline code element since the option we're referring to verbatim has an ellipses.

ahoneybun: It does not follow the style in other articles from my understanding.

storyaddict: We have been moving to `Button` for UI elements, so this is correct.

ahoneybun (approved):
Looks good to me as well.

gangwerz (approved):
Changes look good! Tested live, and everything looks as expected :+1:

PR #981, opened 2022-03-30
restore Disks section and correct QMK link in Launch article

ahoneybun:
This PR does the following:

- Restores the Disks application for Ubuntu and Pop instead of Popsicle as it is installed by default and should be used instead. I need to update the images a bit so this will be a draft at the moment. 
- Fixes a dead link in the Launch article


leviport:
That last commit might fit better in another PR. If you want to include it in this one, please revise the PR title and description to include the QMK link fix.

Ampersandstorm (approved):
I don't see any issues with these updates.

gangwerz (approved):
Looks good!

PR #982, opened 2022-03-30
Replaced GameHub's URL

kutzlect:
The old URL no longer works, so replaced it with the new one.


jacobgkau (approved):
Thank you for updating this link.

leviport (approved):
Thanks for the fix!

ahoneybun (approved):
LGTM :tada:

PR #983, opened 2022-03-30
Update switchable graphics for kudu6

thomas-zimmerman:
Add `kudu6` as a Laptop model that has switchable graphics.
Updated releases on both articles at the same time


leviport (approved):

ahoneybun (approved):
LGTM :tada:

Issue #835, opened 2021-11-12
All Chroot Articles Need Reminder for Internet


storyaddict:
We need to make sure that customers are told to connect their live environment to the internet before copying over their resolv.conf file.

Ampersandstorm:
@storyaddict These are the only two articles I could find mentioning the resolv.conf file. I can add this warning to both, but let me know if there are more: 

https://support.system76.com/articles/pop-recovery/
https://support.system76.com/articles/bootloader/

storyaddict:
There's also this one: https://support.system76.com/articles/login-from-live-disk

PR #989, opened 2022-04-07
internet connectivity before copying over resolv.conf - resolves issue #835

Ampersandstorm:
This PR does the following: 

Adds a note informing the reader to connect to the internet before copying over the /etc/resolv.conf file while in chroot. I've also made some minor wording and punctuation changes.

How should we standardize capitalization of these items?
- systemd (Systemd, SystemD, etc - wikipedia says systemd)
- chroot (I've seen CHROOT and Chroot in our documentation)

### DIFF
--- a/content/login-from-live-disk.md
+++ b/content/login-from-live-disk.md
@@ -21,9 +21,9 @@ tableOfContents: true
 
 ## Login from Live Disk (Chroot)
 
-It is possible to mount an OS drive and log into the installed OS with root access. This is called gaining "chroot" (change to root) access. This process is used for a variety of things. [rescuing files](/articles/disaster-recovery), [fixing package manager issues](/articles/package-manager-pop), [resetting forgotten user passwords](/articles/password), etc.
+It is possible to mount an OS drive and log into the installed OS with root access. This is called gaining "chroot" (change to root) access. This process is useful when [rescuing files](/articles/disaster-recovery), [fixing package manager issues](/articles/package-manager-pop), or [resetting forgotten user passwords](/articles/password).
 
-You'll need a Live OS Environment from which to mount your drive, and log in. This can be done from a [live USB](/articles/live-disk), or, on Pop!_OS, from the [recovery partition](/articles/pop-recovery).
+You'll need a Live OS Environment from which to mount your drive, and log in. This can be done from a [live USB](/articles/live-disk), or on Pop!_OS from the [recovery partition](/articles/pop-recovery).
 Boot the computer while holding down the [boot menu key for your system](/articles/boot-menu), or the <kbd>SPACE</kbd> bar to access Systemd, and the Pop!\_OS Recovery partition.
 
 Once booted into the Live Environment, press <kbd>SUPER</kbd>+<kbd>T</kbd> to open a terminal (Pop!\_OS), or <kbd>CTRL</kbd>+<kbd>ALT</kbd>+<kbd>T</kbd> (Ubuntu) then type this command:
@@ -34,7 +34,7 @@ lsblk
 
 This will show you the name of the main internal drive, which will have 4 partitions (Pop!\_OS), or 3 (Ubuntu) on it.  We will be working with the 3rd partition.  If the main drive is an NVMe drive, it will be called `/dev/nvme0n1p3` (p2 on Ubuntu) and if the drive is a SATA or regular M.2 drive, it will be called `/dev/sda3` (sda2 on Ubuntu).
 
-**NOTE:** The rest of these instructions assume a Pop!\_OS install, for partition labelling. The instructions are otherwise the same. Change the partition number accordingly.
+>**Note:** The rest of these instructions assume a Pop!\_OS install, for partition labelling. The instructions are otherwise the same. Change the partition number accordingly.
 
 Next, run this command:
 
@@ -57,7 +57,7 @@ sudo lvscan
 sudo vgchange -ay
 ```
 
-**Note:** Pay attention to what the `cryptdata` group is called. If it is named something other than `data-root`, substitute the correct info into this next command.  Make sure that `-root` is on the end:
+>**Note:** Pay attention to what the `cryptdata` group is called. If it is named something other than `data-root`, substitute the correct info into this next command.  Make sure that `-root` is on the end:
 
 ```bash
 sudo mount /dev/mapper/data-root /mnt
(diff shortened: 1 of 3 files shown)


jacobgkau:
> How should we standardize capitalization of these items?
> 
> * systemd (Systemd, SystemD, etc - wikipedia says systemd)
> * chroot (I've seen CHROOT and Chroot in our documentation)

For the former, systemd (all lowercase) is the [official branding](https://github.com/systemd/systemd). SystemD is never correct. I could see an argument for using Systemd at the start of a sentence or in title case despite the official branding.

For the latter, chroot (all lowercase) is the name of the case-sensitive command and of the [mechanism](https://en.wikipedia.org/wiki/Chroot). CHROOT is never correct. Once again, I would probably give a pass on capitalizing the first letter at the beginning of a sentence or in title case.

Either way we go for first-letter capitalizations, rewording to avoid using either of these at the beginning of a sentence wherever possible would help to keep the controversy to a minimum.

jacobgkau (commented):
Despite #835, I am seeing that it's not actually necessary to check this-- I can start a chroot using the current instructions (including copying `/etc/resolv.conf`) without being connected to the internet, then plug into Ethernet and use apt, ping, etc. from within the chroot without issue.

Additionally, the `/etc/resolv.conf` file in a live disk (and an installed system) just points to `127.0.0.53`, the local systemd-resolved DNS server. The reason `-n` was added to the `cp` command was to hide an error message: https://github.com/system76/docs/pull/512 But the error message was saying that `/etc/resolv.conf` in the live disk and the chroot were already identical. This means we could really just remove the `cp` command altogether. Testing confirms this.

Separately, we could certainly add a note that if the user normally uses WiFi, they will need to manually connect within the live environment (not at any particular time) because the live environment won't have their WiFi password saved.

content/login-from-live-disk.md
```diff
@@ -34,7 +36,7 @@ lsblk
 
 This will show you the name of the main internal drive, which will have 4 partitions (Pop!\_OS), or 3 (Ubuntu) on it.  We will be working with the 3rd partition.  If the main drive is an NVMe drive, it will be called `/dev/nvme0n1p3` (p2 on Ubuntu) and if the drive is a SATA or regular M.2 drive, it will be called `/dev/sda3` (sda2 on Ubuntu).
 
-**NOTE:** The rest of these instructions assume a Pop!\_OS install, for partition labelling. The instructions are otherwise the same. Change the partition number accordingly.
+>**Note:** The rest of these instructions assume a Pop!\_OS install, for partition labelling. The instructions are otherwise the same. Change the partition number accordingly.
```
storyaddict, line 39: I would reword this as:

The rest of these instructions assume partition labeling consistent with a Pop!\_OS install. For an Ubuntu install, change the partition number accordingly; the instructions are otherwise the same.

content/login-from-live-disk.md
```diff
@@ -46,7 +48,9 @@ If the command fails and says `mount: /mnt: unknown filesystem type 'crypto_LUKS
 
 ### Encrypted Disk
 
-To get access to an encrypted disk, these additional commands need to be run in order to unlock the disk.  Please use the `lsblk` command described above to determine the correct drive and partition.
+To get access to an encrypted disk, these additional commands need to be run in order to unlock the disk.
```
storyaddict, line 51: *To access an encrypted disk, run these additional commands to unlock the encrypted volume.

content/login-from-live-disk.md
```diff
@@ -57,13 +61,13 @@ sudo lvscan
...
 
-**Note:** Pay attention to what the `cryptdata` group is called. If it is named something other than `data-root`, substitute the correct info into this next command.  Make sure that `-root` is on the end:
+>**Note:** Pay attention to what the `cryptdata` group is called. If it is named something other than `data-root`, substitute the correct info into this next command.  Make sure that `-root` is on the end:
 
 ```bash
 sudo mount /dev/mapper/data-root /mnt
 ```
 
-And now the existing hard drive can be accessed by going to the `/mnt` folder.  To use the <u>Files</u> program, go to '+ Other Locations' -> 'Computer' and then click on the `/mnt` folder.
+And now the existing hard drive can be accessed by going to the `/mnt` folder.  To use the <u>Files</u> program, go to `+ Other Locations` -> `Computer` and then click on the `/mnt` folder.
```
storyaddict, line 70: drop the "and"

Now the existing drive can be accessed...

content/pop-recovery.md
```diff
@@ -35,14 +37,13 @@ This option erases the current install along with all user files. It reformats t
 
 Steps to back up user-files from a Live Disk/Recovery can be found [here](https://support.system76.com/articles/disaster-recovery).
 
-**NOTE:**
-The Recovery partition OS version will either be the same as the OS version that shipped with your computer or whichever version to which the Recovery partition has been [updated](#update-recovery-partition).
+>**Note:**The Recovery partition OS version will either be the same as the OS version that shipped with your computer or whichever version to which the Recovery partition has been [updated](#update-recovery-partition).
```
storyaddict, line 40: *or the latest version to which the Recovery partition has been updated

content/pop-recovery.md
```diff
@@ -62,9 +63,9 @@ If the existing install is encrypted, please see the [encrypted disk](#encrypted
 
 If the existing OS install needs to be repaired, the installer application should be closed. An app menu is located in the top-left of the screen with the name of the currently running application (in this case: "Install Pop!\_OS"). Click on the app menu and select `Quit`. Alternatively, you can use the installer app to select keyboard and language settings, and then click the `Try Demo Mode` button in the lower-left corner of the install page.
 
-**NOTE:** be sure not to choose any install or repair options, as this could result in data loss.
+>**Note:** Do not choose any install or repair options, as this could result in data loss.
 
-To access the existing OS drive follow the instructions below.
+For the instructions below to access the existing OS drive:
```
storyaddict, line 68: "For the instructions below" does not make grammatical sense.

content/pop-recovery.md
```diff
@@ -105,7 +106,7 @@ And now the existing hard drive can be accessed by going to the `/mnt` folder.
 
 ## Chroot
 
-`chroot` is the way to run commands as if the existing operating system had been booted.  Once these commands are run, then package manager (`apt`) and other system-level commands can be run.
+`chroot` runs commands as if the existing operating system had been booted.  Once these commands are run, then package manager (`apt`) and other system-level commands can be run.
```
storyaddict, line 109: we don't need the "then" here.

Once these commands are run, package manager...

content/pop-recovery.md
```diff
@@ -156,7 +156,7 @@ Alternatively you can also upgrade the recovery partition from the command-line
 pop-upgrade recovery upgrade from-release
 ```
 
-You should see an output similar to below saying the recovery parition had been updated:
+You should see an output similar to below saying the recovery partition had been updated:
```
storyaddict, line 159: *You should see an output similar to what is pictured below...

OR 

to what is included below

Etc.

storyaddict (changes requested):
Some minor content edits/suggestions.

storyaddict (approved):
Looks good. Thanks for implementing my feedback @Ampersandstorm :)

jacobgkau (approved):

Issue #990, opened 2022-04-10
Multiple broken links in documentation for AMD FSR


pushfoo:
## Page Affected
[content/use-amd-fsr.md](https://github.com/system76/docs/blob/master/content/use-amd-fsr.md) , live at https://support.system76.com/articles/use-amd-fsr/

## What's wrong
All article body links broken, and an image doesn't load.
![image](https://user-images.githubusercontent.com/36696816/162597233-bb0cf436-e9da-4d5d-aeb2-5f56e7651abd.png)

pushfoo:
I think there was a page that's intended to be linked, but isn't being linked correctly: https://github.com/system76/docs/blob/master/content/linux-gaming.md

pushfoo:
Closed by #997

PR #991, opened 2022-04-10
Fixed Broken Link and Image in use-amd-fsr.md

manuelojeda:
[Issue Link](https://github.com/system76/docs/issues/990)


pushfoo:
Although I strongly I agree that a GUI tool may be better for users than a CLI-only tool, recommending an alternative tool is probably best handled as a separate ticket that changes the main linux gaming overview \([`/content/linux-gaming.md`](https://github.com/pushfoo/docs/blob/master/content/linux-gaming.md)\). 

I think the best approach for this issue is prepending `/articles/` to the links on the affected page instead. See [`content/pop-live-disk.md`](https://github.com/system76/docs/blob/master/content/pop-live-disk.md) for an example.

storyaddict:
Addressed in this PR: https://github.com/system76/docs/pull/997

PR #993, opened 2022-04-14
Correcting "Updating on Other OS's" header

cwsystem76:
The plural of the abbreviation "OS" is open to debate, but it's definitely not "OS's", which is a possessive. Changing it to "Updating on Other Operating Systems" which is clearer and obviously correct.


jacobgkau (approved):
I agree with the change rewording for clarity and sidestepping the abbreviation plurality issue.

thomas-zimmerman (approved):
Changes look good.

PR #998, opened 2022-04-25
Pop upgrade 22.04

storyaddict:
- This PR updates wording and steps for updating Pop.
- It also modifies instructions for limited bandwidth to be less specific/urgent.
- Removed "release cycle" article from /docs repo.
- Updates Settings image to reflect new features.


Socialhappiness (approved):
Looks good, great work!

content/upgrade-pop.md
```diff
@@ -87,19 +89,21 @@ sudo apt update
 sudo apt full-upgrade
 ```
 
-Once the updates are applied, a notification should appear at the top of your screen saying that an upgrade is available. Click on this notification, or go to Settings ->  OS Upgrade & Recovery. The System76 upgrade package will display a message that Pop!\_OS 21.10 is available with a `Download` button.
+Once the updates are applied, a notification should appear at the top of your screen saying that an upgrade is available. Click on this notification, or go to Settings ->  OS Upgrade & Recovery. The System76 upgrade package will display a message that Pop!\_OS 22.04 is available with a `Download` button.
```
jacobgkau, line 92: I don't think we refer to this as a `System76 upgrade package` anywhere else. Any reason it can't just say `The upgrade page will display...`?

content/upgrade-pop.md
```diff
@@ -87,19 +89,21 @@ sudo apt update
...
 
 If you are planning on staying on an LTS release for the time being, this is also the page where you can dismiss upgrade notifications.
 
-![Settings OS Upgrade](/images/upgrade-pop/setting-os-upgrade.png)
+The OS Upgrade and Recovery page of settings now allows for automatic update scheduling and updating the Recovery partition as well.
+
+![Settings OS Upgrade](/images/upgrade-pop/Pop-22.04-Upgrade.png)
```
jacobgkau, line 98: It's fine if this filename changes, but the old image should be removed if it's no longer in use.

content/pop-incomplete-upgrade.md
```diff
@@ -22,10 +22,9 @@ section: software-troubleshooting
 tableOfContents: true
 ---
 
-
 ## Repairing Upgrade Errors
 
-Due to the overwhelming demand for 21.10, we have had to expand our servers' bandwidth. If you are experiencing any connection errors please run the following commands in a terminal.
+Due to the overwhelming demand for Pop!\_OS, users may occasionally experience bandwidth issues. If you are still experiencing any connection errors after waiting a short time (15-30 mins.), please run the following commands in a terminal.
```
jacobgkau, line 27: The period followed by a comma around the parenthesis looks weird. I would recommend just spelling out "minutes" to avoid that issue. (That also goes for this sentence in `upgrade-pop.md`.)

content/upgrade-pop.md
```diff
@@ -87,19 +89,21 @@ sudo apt update
 sudo apt full-upgrade
 ```
 
-Once the updates are applied, a notification should appear at the top of your screen saying that an upgrade is available. Click on this notification, or go to Settings ->  OS Upgrade & Recovery. The System76 upgrade package will display a message that Pop!\_OS 21.10 is available with a `Download` button.
+Once the updates are applied, a notification should appear at the top of your screen saying that an upgrade is available. Click on this notification, or go to Settings ->  OS Upgrade & Recovery. The System76 upgrade package will display a message that Pop!\_OS 22.04 is available with a `Download` button.
 
 If you are planning on staying on an LTS release for the time being, this is also the page where you can dismiss upgrade notifications.
 
-![Settings OS Upgrade](/images/upgrade-pop/setting-os-upgrade.png)
+The OS Upgrade and Recovery page of settings now allows for automatic update scheduling and updating the Recovery partition as well.
```
jacobgkau, line 96: I'm wondering how relevant this line is to the upgrade process this article is covering. Mentioning Recovery might not be entirely necessary since that's performed automatically as part of a release upgrade.

content/upgrade-pop.md
```diff
@@ -213,17 +217,17 @@ sudo apt full-upgrade 2>/dev/null | tee ~/upgrade.log
...
 
 ```bash
 pop-upgrade release upgrade
 ```
 
-## Upgrading older releases (20.10)
+## Upgrading older releases to 20.04
 
-Like other older releases, Pop!\_OS 20.10 update files are no longer in the Ubuntu mirror network. Getting your system ready for the upgrade does require some work to manually upgrade the system. Use the following steps to upgrade the system:
+On older releases, Pop!\_OS update files are no longer in the Ubuntu mirror network. Getting your system ready for the upgrade does require some work to manually upgrade the system. Use the following steps to upgrade the system:
```
jacobgkau, line 230: Change from `does require` to `requires`. We're not refuting anything to the contrary, so there's no reason to say `does`.

content/upgrade-pop.md
storyaddict: If the upgrade is failing, updating Recovery and then refreshing will fix the issue.

jacobgkau: That's fair.

jacobgkau: That's fair, if a little out of context here. How about automatic update scheduling, is that relevant to performing a release upgrade?

If we do keep the sentence, could it be combined with the previous sentence into a paragraph? Having "also" in one sentence and then "as well" in the next, on separate lines, makes the section sound long-winded, imo. If this one's following up on the previous one, they should at least be grouped together.

content/upgrade-pop.md
storyaddict: Removed.

content/upgrade-pop.md
storyaddict: Fixed.

content/upgrade-pop.md
storyaddict: Resolved.

content/pop-incomplete-upgrade.md
jacobgkau: Fixed with [bc0c77a](https://github.com/system76/docs/pull/998/commits/bc0c77a6b04b71d92e784ad4c065cffd1033f4b3).

jacobgkau:
Sorry I didn't notice this the first time through, but the screenshot is extremely tall. It takes up the entire height of the browser on a 1080p display, with most of the screenshot being blank.

Screenshot from 2022-04-25 15-55-39

It's also technically outdated since it doesn't include the new scheduler section, which does show up on existing releases. I'm less concerned about the scheduler not being there, but it would be nice if the screenshot was more appropriately sized-- the window just needs to be made vertically smaller before taking the screenshot. I can go ahead and grab one from a VM if it's easier.

storyaddict:
> Sorry I didn't notice this the first time through, but the screenshot is extremely tall. It takes up the entire height of the browser on a 1080p display, with most of the screenshot being blank.
> 
> Screenshot from 2022-04-25 15-55-39
> 
> It's also technically outdated since it doesn't include the new scheduler section, which does show up on existing releases. I'm less concerned about the scheduler not being there, but it would be nice if the screenshot was more appropriately sized-- the window just needs to be made vertically smaller before taking the screenshot. I can go ahead and grab one from a VM if it's easier.

I got one. That was using what Marketing had prefabbed. 

image

jacobgkau:
I grabbed one that doesn't include a non-default active hint color. Here's what it looks like on the page now:

Screenshot from 2022-04-25 16-06-39

content/upgrade-pop.md
```diff
@@ -242,19 +243,19 @@ sudo apt full-upgrade
 sudo apt install pop-desktop
 ```
 
-### 2. Transisiton to the new release files for Pop!\_OS 21.10
+### 2. Transition to the release files for Pop!_OS 20.04
```
jacobgkau, line 246: We were previously recommending the user upgrade directly to 21.10. Has support seen better success using 20.04 as a stepping stone? (If not, it might make sense to update this section to 22.04 instead of lowering it to 20.04.)

jacobgkau: Actually, lowering it to 20.04 means this section no longer covers 20.10, which is what it was covering previously.

Upgrading pre-20.04 releases is already covered on lines 158-222.

storyaddict: You're right. We've got this doubled up now. I removed the 2nd section. Now all releases older than 20.04 get to 20.04, and the newer releases should just upgrade to 22.04 correct?

jacobgkau: It could still be useful to have instructions for upgrading 20.10 to 22.04 manually, since we can no longer upgrade pop-upgrade on that release, and I assume there was a reason that section was added.

I noticed that under the remaining `Upgrading older releases` section, the article recommends doing a refresh or clean install for 20.10 upgrades. That would definitely work, but I don't know why that would be recommended while pre-20.04 upgrades still have manual upgrade instructions available.

jacobgkau (approved):
Like I [said](https://github.com/system76/docs/pull/998#discussion_r858098243), the section for upgrading 20.10 could be added back, but if it's not necessary, that's fine. The rest of the changes look good and the article's in a good state now.

PR #1000, opened 2022-04-27
Update libssl package name for DaVinci Resolve installation

jacobgkau:
The name of this package has changed in 22.04 LTS.

### DIFF
--- a/content/install-davinci-resolve.md
+++ b/content/install-davinci-resolve.md
@@ -24,7 +24,7 @@ DaVinci Resolve requires an NVIDIA GPU to run. (Intel GPUs lack the necessary Op
 The following packages will be needed to build and install DaVinci Resolve:
 
 ```bash
-sudo apt install libssl1.1 ocl-icd-opencl-dev fakeroot xorriso
+sudo apt install libssl3 ocl-icd-opencl-dev fakeroot xorriso
 ```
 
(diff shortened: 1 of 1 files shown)


cwsystem76:
@jacobgkau Is this related to the same `libssl1.1` problem with RStudio as outlined here: https://github.com/system76/careteam/issues/31 ? Or will that require a fix on RStudio's end?

leviport (approved):
I'm seeing libssl3 available in the Jammy repos, but not libssl1.1, so this looks good to me.

jacobgkau:
> @jacobgkau Is this related to the same `libssl1.1` problem with RStudio as outlined here: [system76/careteam#31](https://github.com/system76/careteam/issues/31) ? Or will that require a fix on RStudio's end?

It's related in that both were caused by libssl being updated, but this is a documentation PR for Resolve and that's a packaging issue for RStudio. It sounds like RStudio needs to update their packaging to use `libssl3` as a dependency instead of/alternatively to `libssl1.1`, and they may also need to update their program to work with `libssl3` if they're using any libssl features that changed between 1.1 and 3. Resolve seems to work fine when `libssl3` is installed.

We still need a support approval to merge this Resolve documentation update.

cwsystem76 (approved):
If that's the case, then this looks fine to me. Approved. We'll continue elsewhere regarding the RStudio issue.

PR #1001, opened 2022-04-28
Update 22.04 iso filename

seano-vs:
As the title states.

21.10 -> 22.04

rock on


jacobgkau (approved):
These don't always need to be up-to-date since the article notes they will change (and they change quite often), but no reason not to have it updated for the new release, especially since it's an LTS.

storyaddict (approved):
Nice quality of life change.

PR #1003, opened 2022-04-30
fix missing space in enable-hibernation.md

bentongxyz:
the command at line 79 is missing a single space separating arguments


ahoneybun (approved):
Looks good!

PR #1006, opened 2022-05-06
DisplayLink notes and more

ahoneybun:
This PR does the following:

- Adds note about galp5-3050 USB Type-C charging (does not work as spec only goes up to 100W and it needs 120W)
- Adds note about DisplayLink not supporting 5.15+ kernel versions
- Updates DisplayLink version


jacobgkau (changes requested):
Sales page tech specs and tech docs both say galp5-3050 can charge with USB-C.

image

image

Where are you getting the information that it's not?

jacobgkau:
> It can charge but try to find a docking station that outputs over 100W.

This table provides no context about only being about docking stations. You could add wattage requirements for each model instead, but as it is now, this is adding incorrect information.

ahoneybun:
This is an article just about docking stations so I don't think it's incorrect information.

jacobgkau:
> IIRC, we have not confirmed that the galp5-3050 would charge well with a charger over 100W. 100W was the max supported at the time the galp5-3050 was made. 200W USB-C charging is recently on the market, and I think we should confirm the system can pull more than 100W on that port before we say it's supported

We already say it's supported, so we are beyond that point. Also, we did confirm it charges at a slower rate with a <100W charger. Please work with the appropriate parties to get a larger charger ordered for testing, in the meantime, we are not going to merge a PR conflicting with two of our other sources of information plus the upstream manufacturer's documentation on a whim.

ahoneybun:
@jacobgkau if I remove the note about the USB-C charging would you approve it at that point?

storyaddict (changes requested):
Remove the 3050 in the table, and I'm fine with the changes.

storyaddict (approved):
Changes look good.

gangwerz (approved):
Changes looks good!

I removed the galp5-3050 reference from this PR's title for clarity, as there are no longer changes related to that system.

PR #1008, opened 2022-05-12
Expanding the repo management doc, new Flatpak command line doc

cwsystem76:
This expands the "Manage Repositories" support doc, adding new screenshots and more details about using `repoman`. The original doc contained a section on Flatpak commands that were unrelated to repo/source management, so I broke those out in an expanded "Using Flatpak from the Command Line" doc with its own page. I also updated the keywords for both to include more recent Pop!OS versions and the "Flatpak" keyword.

Setting this up as a draft as I haven't made this level of changes to the `docs` repo before on a single PR.


content/manage-repos-pop.md
```diff
@@ -17,70 +24,62 @@ section: software
...
-flatpak run org.inkscape.Inkscape
+flatpak remotes
 ```
 
-This command will run inkscape once it is installed on the system.
+To add a new Flatpak source (with Flathub as the example):
 
 ```bash
-flatpak update
+flatpak remote-add --if-not-exists flathub https://dl.flathub.org/repo/flathub.flatpakrepo
```
thomas-zimmerman, line 76: This should have `--user` to add a flathub remote source as a user repository.

thomas-zimmerman (changes requested):
Changes with the screen shots look good. Would like to get the update for the suggested command to be a bit safer with `--user` repositories.

content/manage-repos-pop.md
cwsystem76: That's how it was on the original document. I've gone ahead and added `--user` to both the `remote-add` and `remote-delete` commands for consistency.

thomas-zimmerman (approved):
Nice updates.

content/manage-repos-pop.md
```diff
@@ -17,70 +24,62 @@ section: software
...
+## Repositories and Sources
 
-Pop!_OS' tool for adding, removing and editing repositiories is called <u>Repoman</u> which is a part of the Pop!_Shop.
+Pop!\_OS comes pre-installed with a number of sources for installing additional software, but additional sources can be added as well, either by hand or by installing a third-party package. Sources for systemwide software packages are called _repositories_ (or "repos" for short), and sources for Flatpak software are called just _sources_.
 
-![Pop!_OS Shop](/images/manage-repos/pop-shop-button.png)
+## Repoman and System Repositories
 
-Click on the gear icon in the top right of the <u>Pop!_Shop</u> window. Once <u>Repoman</u> is opened, click on the **Extra Sources** tab there will be options to add repositories with the **Plus** button (on the bottom left). Select a repository from the list and then click on the **Edit** button (next to the **Plus** button) to edit the selected repository.
+Pop!\_OS' tool for adding, removing and editing repositiories and sources is called <u>Repoman</u>, and it is accessible through the Pop!_Shop.
```
Ampersandstorm, line 33: This would be "Pop!_OS's" unless "Pop!_OS" is a plural noun.

cwsystem76: Oops. Fixed.

cwsystem76:
> Looks good overall. Do you think it would be beneficial to describe PPA's and explain why someone might want to be cautious about adding them?

That might be useful, but I think it can wait for a future update, as I think it would require significantly more tuning, so to speak.

Ampersandstorm (approved):

PR #1009, opened 2022-05-13
Updated Bluetooth to `pipewire-pulse`

MendyBerger:
Replaced `pulseaudio-module-bluetooth` with `pipewire-pulse` as I understand that this is used since v22.04.

Someone with a bit more knowledge in this area should probably make sure that this is actually the case


mmstick:
Bluetooth support for pipewire is managed by `wireplumber`

ahoneybun (approved):
LGTM and work on my Pop 22.04 install.

Ampersandstorm (approved):
Looks good.

PR #1010, opened 2022-05-18
Update article with new commands

ahoneybun:
This PR does the following. 

- Update commands
- Add link to Ubuntu docs


content/containers.md
```diff
@@ -30,26 +30,37 @@ Ubuntu : <kbd>CTRL</kbd>+<kbd>ALT</kbd>+<kbd>T</kbd>
...
+snap install lxd
 ```
 
 ## Setup
 
 Tell LXD what storage and network is needed with this command:
 
 ```bash
-sudo lxd init
+lxd init
```
jacobgkau, line 42: I had to add my user to the `lxd` group and reboot before these commands would work (should be doable with `sudo usermod -aG lxd $USER`.) Before that, the first one (init) didn't complete properly and the second one (launch) just gave me a "permission denied" error.

ahoneybun: Should be fixed in the newest commit.

jacobgkau: The reboot that I mentioned was not optional, your updated instruction still don't work. I'll add it myself.

jacobgkau (approved):

sebastian-burlacu (changes requested):
Line 30 would be better as "Install the software by running these commands in the terminal:"

jacobgkau:
> Line 30 would be better as "Install the software by running these commands in the terminal:"

You're right, there are two commands so "this command" was incorrect. Fixed with 2abd367.

PR #1013, opened 2022-05-28
Update Bluetooth troubleshooting commands for v22.04

cwsystem76:
Some of the instructions on this document no longer apply for Pop!OS v22.04, so I've split up those commands into a version for v22.04 and a verison for previous releases. I also added a note to the section about the `module-bluetooth-discover` Pulseaudio module, which is no longer used in v22.04.


ahoneybun (changes requested):
We would need to update the Pop name to Pop!_OS you can use \ to escape the name for formatting.

cwsystem76:
> We would need to update the Pop name to Pop!_OS you can use \ to escape the name for formatting.

Done.

storyaddict (approved):
These changes make sense and look correct.

Ampersandstorm (approved):
Looks good!

PR #1016, opened 2022-06-09
Update wording and commands for Arch users

thomas-zimmerman:
Commands were not quiet clear that you could install with Paru *or* directly.


content/system76-driver.md
```diff
@@ -97,19 +97,20 @@ makepkg -si
 
 **NOTE** that by default <u>Paru</u> uses vim standards so when you see a ":" press the <kbd>q</kbd> key to continue. You may also need to import some public keys by using the <kbd>y</kbd> key.
 
-Now the <u>System76 Driver</u> can be cloned, built and installed using these commands:
+Next, build the <u>System76 Driver</u> with <u>Paru</u>
```
jacobgkau, line 100: This sentence is missing punctuation.

content/system76-driver.md
```diff
@@ -97,19 +97,20 @@ makepkg -si
...
+**NOTE** choose the first software option listed after running the <u>Paru</u> command.
```
jacobgkau, line 101: We have two all-caps bold **NOTES** within three paragraphs of each other. This is an instruction, can it not just be stated plainly?

If you did keep the **NOTE**, it needs a colon after it in this context. (It looks like that earlier one doesn't have a colon because it's part of the sentence, which I would not have agreed to, but that doesn't work here.)

content/system76-driver.md
```diff
@@ -97,19 +97,20 @@ makepkg -si
...
 ```bash
-git clone https://aur.archlinux.org/system76-driver.git
-cd system76-driver
-makepkg -srcif
+paru -s system76-driver
 sudo systemctl enable --now system76
 ```
 
-**NOTE** choose the first software option listed after running the <u>Paru</u> command.
+If you would like to build the package manualy, use the following commands:
```
jacobgkau, line 108: This is failing lint because of the trailing space.

jacobgkau (changes requested):
I don't really like that we have the recommended instructions installing a specific AUR helper (with AUR helpers being unsupported by Arch and this being just one of many), but I agree with this minor reorganization for clarity.

There are some grammatical and style issues with just the lines being changed, which I've notated.

After speaking with @Ampersandstorm, we'd prefer the word "Note" not be all-caps when it's used like that. Just **Note:** followed by the note, in a quote block (>) to visually separate it from the surrounding paragraphs, is sufficient and looks nicer.

thomas-zimmerman:
Updated PR in https://github.com/system76/docs/pull/1016/commits/ad5a83094b86f0314ecfba871a103a8f8df0e8c9 with requested Changes.

ahoneybun:
> I don't really like that we have the recommended instructions installing a specific AUR helper (with AUR helpers being unsupported by Arch and this being just one of many), but I agree with this minor reorganization for clarity.

We'll want to update it if `paru` is no longer being updated but I think recommending a helper isn't an issue and really anything from the AUR is unsupported.

ahoneybun (changes requested):
Maybe we move the manual way to build the package right under installing the build deps then list Paru if they want to use a helper?

jacobgkau (approved):
I'm approving because the current wording is confusing/incorrect (involves installing twice with no distinction) and the current changes make this more sane. However, I agree with @ahoneybun's suggestion that it would be better to list the standard method of installation first, then give the AUR helper example second, if anyone would like to make that change.

storyaddict (approved):

PR #1017, opened 2022-06-12
Create enable-wayland.md

uncivilgentleman:
I realize this isn't great documentation but it's hopefully better than nothing.


jacobgkau (changes requested):
Wayland is disabled by default for a reason (it does not work as expected on all System76 hardware, and it is therefore not supported for the time being.) At the very least, an article like this would need to explain that this is unsupported (maybe with some examples of problems that can show up), and should include instruction for reverting back to Xorg. We may also want to consider making this be part of the Community Articles section since it is not something we want to endorse right now.

storyaddict (dismissed):
Approving this for the text. I'd like screenshots to be added in another PR after. 
I took some screenshots in a VM which I can add.

content/enable-wayland.md
```diff
@@ -0,0 +1,20 @@
...
+ - Display Manager
+
+hidden: false
+section: software
+tableOfContents: true
+---
+
+**Note:** *Using Wayland is not supported on Pop!\_OS and may cause problems particularly when using Nvidia graphics cards. Do so at your own risk.*
+
+Edit `/etc/gdm3/custom.conf` and comment out the line `WaylandEnable=false`. Save the file and restart your computer.
```
jacobgkau, line 16: This doesn't explain what "commenting out" means, or how to edit this file with elevated privileges. These instructions are not sufficient.

jacobgkau (changes requested):

ahoneybun:
@uncivilgentleman I'm closing this issue as it is stale but please do let us know if you have any questions about updating.

PR #1018, opened 2022-06-15
Add commands to clear wireplumber saved settings

thomas-zimmerman:
expand section to clear current saved settings for PulseAudio and Wireplumber

### DIFF
--- a/content/audio.md
+++ b/content/audio.md
@@ -153,6 +153,8 @@ rm -r ~/.config/pulse/*
 rm -r ~/.local/state/wireplumber/*
 ```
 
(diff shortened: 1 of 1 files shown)


cwsystem76:
There may be cases where those two commands come back with `rm: cannot remove '/home/ckw/.config/pulse/*': No such file or directory` or similar, and it's probably worth clarifying that an error like that is no reason to worry.

Also, we should be clear that they need to reboot after running those commands to make sure that the settings are reset. If they run them and just keep using the system, things might not work as expected until they do restart.

n3m0-22 (approved):
Looks good.

PR #1021, opened 2022-06-17
Move Paru commands after installing from source commands for arch.

thomas-zimmerman:
Include why you would want to use a AUR helper and move it below the direct build instructions.


ahoneybun (approved):
Looks good to me and makes it more clear that you do not need to use a helper.

PR #1027, opened 2022-07-05
fix(enable-hibernation): clarify vulnerabilities

qin-guan:
The security benefits were not further elaborated on and the following sentence makes the statement confusing as it is actually talking about vulnerabilities in hibernation


jacobgkau:
If you wish to make this wording less confusing, you could get rid of the first sentence as you've already done, keep "Additionally" as the first word instead of "However," and add another sentence or clause stating that hibernation eliminates this risk by not keeping the disk unlocked.

qin-guan:
My apologies, I overlooked the "suspend" in the comparison... Thanks for the feedback though! I've updated the content 😃

jacobgkau (approved):
The meaning is now preserved, and the new wording seems more clear and direct to me than the old wording. The new wording is also more positive.

PR #1029, opened 2022-07-09
system76-power not starting

fnatical6:
`system76-power.service` may not start on boot even though it's been enabled as per [this](https://github.com/pop-os/system76-power/issues/299#issuecomment-1033022670) explanation.

This in turn can cause the System76 Power GNOME Shell Extension to fail to load within the Extensions application.


content/system76-software.md
```diff
@@ -222,6 +228,14 @@ make install
 
 Now log out and use the <u>Extensions</u> to enable the extenstion.
 
+If `Gio.DBusError` presents itself when attempting to enable the extension within the Extensions application, confirm that the `system76-power.service` is running:
+
+```bash
+sudo systemctl status system76-power.service
+```
+If the service isn't running, revisit [these](https://github.com/system76/docs/edit/master/content/system76-software.md#system76-power-gnome-shell-extension-in-fedora) steps.
```
Ampersandstorm, line 236: Thanks for adding this! Although we'd typically link using the relative path: 

`[these](/articles/system76-software.md#system76-power-gnome-shell-extension-in-fedora)`

However, I'm wondering if this link is necessary since is jumps to a section that will likely still be visible to the reader. Could we move those instructions down here? 

Alternatively we could say:

`If the service isn't running, revisit the above instructions to mask the `power-profiles-deamon.service`.

I can implement either of these suggestions, but let me know if you have any concerns.

Ampersandstorm (changes requested):
Instructions to change to relative link, suggest moving masking instructions below line 236.

fnatical6 (commented):
Sounds good!

ahoneybun (approved):

fnatical6:
Those suggestions sound perfect thanks David!

PR #1034, opened 2022-07-20
Strengthening Disclaimer

storyaddict:
This PR Does the Following:

- Combines the "scope of support" section with the "disclaimer section." 
- Clarifies and strengthens the language around what System76 will/won't support.
- Gives an example of the kind of problems user may expect to encounter when running Windows on System76 hardware.


cwsystem76:
While you're working on this, would it be possible to also update the Windows 11 section? It talks about TPM support, but not Secure Boot, which is the real sticking point. Our Coreboot systems have a TPM but not Secure Boot, and it's the latter that's the real issue.

storyaddict:
> While you're working on this, would it be possible to also update the Windows 11 section? It talks about TPM support, but not Secure Boot, which is the real sticking point. Our Coreboot systems have a TPM but not Secure Boot, and it's the latter that's the real issue.

I'm giving this a second pass to make those changes now.

cwsystem76:
> Open Firmware systems to do support Secure Boot which is a requirement for Windows 11.

I think this is meant to be "do not support"?

content/windows.md
```diff
@@ -28,15 +28,18 @@ section: software
...
 ---
 
-## Disclaimer
+## Important Disclaimer - Please Read
 
 System76 is not a licensed reseller or installer of the Windows operating system. This article is provided for information purposes only.
 
-System76 encourages users to take ownership of their machines and install whatever software or operating systems they prefer. However, System76 does not guarantee the success or quality of experience when installing Windows.
+System76 encourages users to take ownership of their machines and install whatever software or operating systems they prefer. However, System76 does not guarantee the success or quality of experience when installing Windows and does not provide support or testing for any Windows version. Please do not open a support ticket for anything other than clarification of the steps covered in this article.
+The contents of this support article are the **total extent of support and troubleshooting that System76 can provide for Windows.** Any troubleshooting or Windows support questions not covered in this article are outside the scope of System76 and should be referred to Microsoft.
```
Ampersandstorm, line 36: Consider changing `Any troubleshooting or Windows support questions not covered in this article are outside the scope of System76 and should be referred to Microsoft.`

to:

`Any Windows-related support for System76 hardware not covered in this article is beyond the scope of support provided by System76. Please refer to Microsoft's [official troubleshooting documentation](https://support.microsoft.com/en-us) or other online resources.`

Ampersandstorm (changes requested):
I'm fine with these changes as long as this disclaimer doesn't conflict with statements in System76 marketing material, and falls in-line with Emma and Sam's support expectations. We don't want people thinking some warranty has been voided just because they tried to install Windows and now their system won't startup for whatever reason. I wonder if adding steps to restore Pop!_OS (or a link to an article showing how to install Pop!_OS) would be helpful.

ahoneybun (approved):
LGTM

n3m0-22 (changes requested):
The new changes look good. The three changes suggested still need to be addressed.

n3m0-22 (approved):
Looks good.

PR #1035, opened 2022-07-27
Add oryp9, lemp11, darp8, and gaze17

ahoneybun:
This adds these models to the list of Open Firmware and Open EC per @crawfxrd .


jacobgkau (approved):

thomas-zimmerman (approved):
LGTM

PR #1039, opened 2022-08-10
Add launch Lite

ahoneybun:


content/launch-keyboard.md
```diff
@@ -32,7 +29,11 @@ The core idea of "Layers" is that one key can be used for different actions. For
 
 ## Firmware Updates
 
-Firmware updates for the Launch are handled using [fwupd](https://fwupd.org/) and our <u>Firmware Manager</u> can be used to apply the update. When you get a notification about a firmware update, you will need to unlock the keyboard using Fn+Esc (**Note:** that these will be the default keys even if you remapped them in the Keyboard Configurator) before you can apply the update. When the keyboard is unlocked, you will see the following keys light up:
+Firmware updates for the Launch and Launch Lite are handled using [fwupd](https://fwupd.org/) and our <u>Firmware Manager</u> (which can be installed on Arch and Fedora using this [article](https://support.system76.com/articles/system76-driver/)) can be used to apply the update. When you get a notification about a firmware update, you will need to unlock the keyboard using <kbd>Fn</kbd> + <kbd>Esc</kbd> (Launch Lite is <kbd>Left Ctrl</kbd> + <kbd>Right Ctrl</kbd> + <kbd>Esc</kbd>).
+
+> ℹ️ These will be the default keys (even if you remapped them in the Keyboard Configurator) that must be used before you can apply the update.
```
leviport, line 34: >(even if you remapped them in the Keyboard Configurator)

Should this be clarified? I could see readers being uncertain whether this means that they need to use the _physical_ keys in these locations, or they need to push left and right ctrl plus escape, no matter where on the board they have been moved to. The latter is true btw. 

They can also map the Reset key somewhere on the keyboard and use that instead, just in case they don't want a right ctrl key or something.

Ampersandstorm: I wasn't sure about this either, and I think I'm still not clear, would this be correct?

> ℹ️ If you remapped the <kbd>Fn</kbd>, <kbd>Esc</kbd>, <kbd>Left Ctrl</kbd>, or <kbd>Right Ctrl</kbd>, you will need to press whichever keys now correspond with your new mapping.

thomas-zimmerman (approved):
This looks good. Looks like it resolves any questions that @leviport had on this update.

leviport (approved):

PR #1045, opened 2022-08-22
Remove superfluous sentence under troubleshooting

Pantalon-Capitaine-Fromage:
"Troubleshooting

Most upgrades proceed without a hitch, but occasionally things go wrong. If your system hangs on the upgrade process or if it reboots and won't start, then these steps can help repair your broken OS.

Most upgrades proceed without issues, but occasionally things go wrong. If your system hangs on the upgrade process or if it reboots and won't start, then these steps can help repair your broken OS."


leviport (approved):
Thanks for the fix!

Ampersandstorm (approved):
Good catch, thanks!

PR #1049, opened 2022-08-26
Update desktop-environment.md

smutao:


content/desktop-environment.md
```diff
@@ -43,6 +43,8 @@ If multiple desktop environments are installed, GDM will display a gear icon, wh
 
 ![Login2](/images/desktop-environment/Login2.png)
 
+If you do not find the gear icon to switch the desktop environment, click the user icon first to see the gear button.
```
jacobgkau, line 46: "Do not find" isn't a very natural English expression. It should either be "do not see" ("don't see") or "can't find."

jacobgkau (commented):
On a system with only a single DE available, clicking the user icon does not display a gear icon. Is this just telling the reader that they need to select their user before the menu will become available?

content/desktop-environment.md
smutao: I agree. Thanks.

jacobgkau:
Can you answer the other question?

> On a system with only a single DE available, clicking the user icon does not display a gear icon. Is this just telling the reader that they need to select their user before the menu will become available?

This wording could mislead users when they only have a single desktop environment installed or available (e.g. after installing a Wayland-only DE, they'll still only have one DE available because Wayland is disabled out of the box.) Clicking on the user icon will not make the gear appear.

The more appropriate place to clarify this would be higher up before referring to the gear icon. "If multiple desktop environments are installed, after selecting your user on the login screen, GDM will display a gear icon, which will allow you to select the desktop environment you want to launch."

The screenshot for this section is also outdated; the gear icon should be in the bottom right of the screen, not next to the Sign In button. I can grab an updated screenshot.

smutao:
> Can you answer the other question?
> 
> > On a system with only a single DE available, clicking the user icon does not display a gear icon. Is this just telling the reader that they need to select their user before the menu will become available?
> 
> This wording could mislead users when they only have a single desktop environment installed or available (e.g. after installing a Wayland-only DE, they'll still only have one DE available because Wayland is disabled out of the box.) Clicking on the user icon will not make the gear appear.
> 
> The more appropriate place to clarify this would be higher up before referring to the gear icon. "If multiple desktop environments are installed, after selecting your user on the login screen, GDM will display a gear icon, which will allow you to select the desktop environment you want to launch."
> 
> The screenshot for this section is also outdated; the gear icon should be in the bottom right of the screen, not next to the Sign In button. I can grab an updated screenshot.

Yes, that was my original idea "to select user first before seeing the gear icon". Your sentence is much better and more accurate. Thanks a lot.

PR #1051, opened 2022-08-31
Update bluetooth.md

staf0048:
Updated Bluetooth Troubleshooting section to explain how to eliminate the possibility that Bluetooth issues are related to hardware.

### DIFF
--- a/content/bluetooth.md
+++ b/content/bluetooth.md
@@ -70,9 +70,9 @@ This process lowers the sound quality of the stream when in HSP/HFP mode, so aud
 
 ## Bluetooth Troubleshooting
 
-Bluetooth issues can be troubleshooted in several ways.  The first thing to check is toggling airplane mode which will sometimes get Bluetooth functioning again.  Next, make sure Bluetooth is enabled in the top bar, or in the <u>Bluetooth</u> system settings.
+Bluetooth issues can be troubleshooted in several ways.  The first thing to check is toggling airplane mode which will sometimes get Bluetooth functioning again.  Next, make sure Bluetooth is enabled in the top bar, or in the <u>Bluetooth</u> system settings.  In addition, sometimes your motherboard may not recognize your Bluetooth controller after resuming from suspend or a reboot.  Eliminate this possibility before making any of the software changes described below by shutting down, disconnecting the power, and holding the power button for 30 seconds to discharge your motherboard's capacitors.  Then, reconnect the power, boot the machine, and see if Bluetooth is working again.
 
-Then, try reinstalling Bluetooth related software with this command, depending on the verison of Pop!\_OS you're using.
+If the above does not resolve the issue, then try reinstalling Bluetooth related software with this command, depending on the verison of Pop!\_OS you're using.
 
 *For Pop!\_OS 22.04 or higher:*
 


content/bluetooth.md
```diff
@@ -70,9 +70,9 @@ This process lowers the sound quality of the stream when in HSP/HFP mode, so aud
 
 ## Bluetooth Troubleshooting
 
-Bluetooth issues can be troubleshooted in several ways.  The first thing to check is toggling airplane mode which will sometimes get Bluetooth functioning again.  Next, make sure Bluetooth is enabled in the top bar, or in the <u>Bluetooth</u> system settings.
+Bluetooth issues can be troubleshooted in several ways.  The first thing to check is toggling airplane mode which will sometimes get Bluetooth functioning again.  Next, make sure Bluetooth is enabled in the top bar, or in the <u>Bluetooth</u> system settings.  In addition, sometimes your motherboard may not recognize your Bluetooth controller after resuming from suspend or a reboot.  Eliminate this possibility before making any of the software changes described below by shutting down, disconnecting the power, and holding the power button for 30 seconds to discharge your motherboard's capacitors.  Then, reconnect the power, boot the machine, and see if Bluetooth is working again.
```
leviport, line 73: This would only apply to desktops, and not many laptops have (easily) removable batteries these days. It also sounds like overkill. What if we just told users to fully power off the machine, wait a bit, then power it back on with the power button? The motherboard will have to re-initialize PCIe and USB, which would re-initialize any bluetooth adapters that may be present.

staf0048: I'm not certain if a shutdown and wait approach would actually discharge the capacitors quickly enough.  Eventually it would work, but we'd need to provide a reasonable time frame.  In my case though, I did try to shutdown (also unplugged my machine), then waited 5 minutes before rebooting.  That didn't have any impact.  I ended up having to do the same, but also hold the power button to clear out any remaining voltage before rebooting.

That said, I believe holding the power button is sufficient whether the machine is plugged in to the wall or connected via battery to discharge the capacitors quickly.  So a more universal approach could be to simply shut down, press and hold the power button, then reboot to see if bluetooth is working again.

leviport: I'm still not certain discharging the capacitors is something that has to be done to fix bluetooth for the vast majority of machines. Is this something you have to do on your machine regularly?

staf0048: It's a new machine that I built last week and have experienced the same issue twice so far.  Both times after waking from suspend after putting it to sleep for the night.

I completely agree that it may not affect a majority of users, but I think it's an important and simple step to try 1st before doing any of the software changes described in the rest of article.  The reason I submitted is because I tried those changes myself (along with several other solutions from various sources) to no avail, but this simple trick worked.  Hoping I can help save another person from hours of research and testing different solutions, only to end up getting frustrated that all their work did nothing and they end up dropping pop_os or gnu/linux altogether because they couldn't figure out Bluetooth - when all they needed to do was hold the power button for 30 seconds.  :-)

To be clear though, this is specific to anyone who had a previously working Bluetooth that has suddenly stopped.  If Bluetooth had never worked, then this solution won't solve that.  In either case though, it is still a non-destructive step to take to ensure there aren't any hardware issues before users start messing around with config files, deleting files/folders, installing other software, etc.

staf0048: FYI - Another user impacted for which this fix worked:

https://www.reddit.com/r/pop_os/comments/xsh8w3/comment/iqri222/?context=3

leviport:
Thanks again for the contribution, but I think this one is enough of a rarity that it probably doesn't fit very well in these docs.

PR #1054, opened 2022-09-03
Update Yubikey article

thomas-zimmerman:
This looks like a better state of setting up Yubikey. I found that my key does not function correctly to work here but the software side looks correct to me now. 

I'd really like someone with a working Yubikey to test that these steps lead to a working setup.


ahoneybun (approved):
Steps worked for me and feel more clear then before.

content/yubikey-login.md
```diff
@@ -15,98 +15,94 @@ tableOfContents: true
...
 - /etc/pam.d/login
 - /etc/pam.d/gdm-password
 
-### /etc/pam.d/common-auth
-
-At a minimum you need to modify this file.
+## Debugging and Testing your PAM configuration
 
-Add the auth line to the top of this file (after the comment block).
+You will want to keep a root terminal logged in while setting this up to make sure you can back out any changes that do not allow you to login.  In the initial setup, the parameters to setup libpam-yubico included debug message. "Sudo" on the command line will now show debug output when run.
```
leviport, line 90: >make sure you can back out any changes

Should probably change "back out any changes" to "reverse any changes" for clarity?

content/yubikey-login.md
```diff
@@ -15,98 +15,94 @@ tableOfContents: true
...
+    sudo dpkg-reconfigure libpam-yubico
+
+You will want to change the "Parameters for Yubico PAM:" to be:
+
+    mode=challenge-response debug chalresp_path=/var/yubico
+
+![dpkg-reconfigure-pg1](/images/yubikey-login/dpkg-reconfigure-pg1.png)
+![dpkg-reconfigure-pg2](/images/yubikey-login/dpkg-reconfigure-pg2.png)
+
+If you want to only want to add the Yubikey as sufficient, you will need to change the second setting in the following line in ```/etc/pam.d/common-auth```:
```
leviport, line 72: Something makes me feel like "sufficient" should be encouraged, so I wonder if we shouldn't make this instruction sound stronger than an "if you want to". I imagine that it'll make it less likely for users to lock themselves out.

content/yubikey-login.md
```diff
@@ -15,98 +15,94 @@ tableOfContents: true
...
+You will want to keep a root terminal logged in while setting this up to make sure you can back out any changes that do not allow you to login.  In the initial setup, the parameters to setup libpam-yubico included debug message. "Sudo" on the command line will now show debug output when run.
 
-You may need to run `pam-auth-update` afterwards.
+With the Yubikey plugged into the system simulate a login with the following commands:
 
-### /etc/pam.d/login
+support@pop-os:~$ sudo -s
+[snip lots of logging]
+[sudo] password for support:
+root@pop-os:/home/support#
```
leviport, line 97: This should be in code blocks

content/yubikey-login.md
```diff
@@ -15,98 +15,94 @@ tableOfContents: true
...
 ## Configure Plugable Authentication Modules
 
 **Before making any changes to the files listed below, I highly recommend backing up each file, and having a sudo/root session open in case you need to roll-back.**
 
-You need to add the following line to each of the files listed below:
+    sudo dpkg-reconfigure libpam-yubico
+
+You will want to change the "Parameters for Yubico PAM:" to be:
+
+    mode=challenge-response debug chalresp_path=/var/yubico
```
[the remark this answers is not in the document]
leviport: Ah, I see, this is covered later. I wonder if a note about the `debug` setting, with a link to the "debugging and testing" section, would be appropriate here, just in case readers have a short attention span like I do.

content/yubikey-login.md
ahoneybun: Code blocks broke the lint as it read it as HTML I believe.

leviport: Well it currently looks like this:



I want it to not look like that.

content/yubikey-login.md
```diff
@@ -15,98 +15,94 @@ tableOfContents: true
...
-To enable challenge-response on your Yubikey, type the following command:
+To enable challenge-response on your Yubikey in slot 2, type the following command:
 
-    ykpersonalize -2 -ochal-resp -ochal-hmac -ohmac-lt64 -oserial-api-visible
+    ykman otp chalresp -g 2
 
 This configures slot 2 for challenge-response, and leaves slot 1 alone.
 
-Next we need to create some challenge response files, move them to a system wide directory, and secure those files:
+Next we need to create a place to sore your challenge response files, secure those files, and finally create the stored challenge files:
```
leviport, line 38: s/sore/store

ahoneybun (approved):

content/yubikey-login.md
leviport: This wasn't resolved.

ahoneybun: Should be better now.

thomas-zimmerman: Do we think that we should change the suggested state to be "sufficient" and only recommend that you switch to "required" after testing?

content/yubikey-login.md
```diff
@@ -15,98 +15,96 @@ tableOfContents: true
...
-You should receive a unique *challenge-123456* in your output.
+You should receive a unique *challenge-serial* in your output.
 
 Now, to finish up:
 
     sudo mv ~/.yubico/challenge-123456 /var/yubico/alice-123456
-    sudo chown root.root /var/yubico/alice-123456
-    sudo chmod 600 /var/yubico/alice-123456
+    sudo chown root.root /var/yubico/bob-123456
+    sudo chmod 600 /var/yubico/lisa-123456
```
leviport, line 55: I'm confused by the usage of `alice` then `bob` then `lisa` in these examples. When I set my yubikey up, will I be changing the names in each command as I'm running them, like these examples show? If not, we should use only one name and keep it the same every time.

Also, further up, `123456` was changed to `serial` but it is still `123456` in these examples. Should that also be made more consistent?

jojo2357:
I have a yubikey bio, and this tutorial nor the old tutorial work for me

```
jojo2357@pop-os:~$ ykman config usb -e OTP
WARNING: PC/SC not available. Smart card protocols will not function.
Error: OTP not supported over USB on this YubiKey.
```

ahoneybun:
> I have a yubikey bio, and this tutorial nor the old tutorial work for me
> 
> ```
> jojo2357@pop-os:~$ ykman config usb -e OTP
> WARNING: PC/SC not available. Smart card protocols will not function.
> Error: OTP not supported over USB on this YubiKey.
> ```

This was done with USB-A and USB-C Yubikeys so I'm not sure of your model.

leviport (approved):

jojo2357:
> ...I'm not sure of your model

This one: https://www.yubico.com/product/yubikey-bio-series/yubikey-c-bio/

the usb-c variant. Need I mail my key to you for testing? It isnt very useful to me right now \:(

ahoneybun (approved):
Works well and better steps now.

ahoneybun:
> > ...I'm not sure of your model
> 
> This one: https://www.yubico.com/product/yubikey-bio-series/yubikey-c-bio/
> 
> the usb-c variant. Need I mail my key to you for testing? It isnt very useful to me right now :(

This is the one that I have for testing that I use myself:

https://www.yubico.com/product/yubikey-5-series/yubikey-5c/

I'm not sure about doing that testing.

leviport:
I think the bio one is a different thing entirely, since it has a fingerprint scanner. I don't believe this article was ever intended to apply to the bio ones.

PR #1056, opened 2022-09-09
feat: Update DAC crackling instructions for PipeWire

### LINKED ISSUE #1044: Audio crackling fix doesn't work anymore for 22.04
Hello,
I used [this troubleshooting step](https://github.com/system76/docs/blob/0373a5b7a3597ad0a9ae3a77908f7ca2945436f2/content/audio.md#audio-crackling-or-hardware-clicking) to fix issues with my wireless headphones waking up too slow and cutting off the first few seconds of audio. After upgrading to 22.04 this fix doesn't work anymore, probably because Pop!_OS uses pipewire now so the pulse configuration files are ignored. Would it be possible to add instructions on how to fix this issue on 22.04 and onward?

jacobgkau:
Closes #1044.

Adds the new instructions for PipeWire, but keeps the old ones for PulseAudio around for reference.

The new PipeWire instructions have been confirmed by several users in Mattermost, and I've also tested it myself on my serw10.

### DIFF
--- a/content/audio.md
+++ b/content/audio.md
@@ -159,7 +159,7 @@ rm -r ~/.local/state/wireplumber/*
 
 If you hear audio crackling (especially when you start or stop playing audio), your audio card may be going to sleep too often. This is known to happen on some versions of the [Serval WS](/articles/serval-dac/) and some [Thunderbolt docks](https://github.com/system76/docs/issues/491).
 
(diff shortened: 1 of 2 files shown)


Ampersandstorm (approved):
Looks good, thanks for updating

leviport (approved):

PR #1064, opened 2022-09-22
Update for Plugable UD-ULTC4K

jlangson:
It does not natively support multiple monitors with the Lemur. I tested this myself and when it didn't work, I confirmed this with Pluggable's  support.

Apologies for not linking to the PR number, but I don't know what the number will be before the PR is made.


n3m0-22 (dismissed):
Looks good. Thank you.

jacobgkau:
@jlangson What specific generation of Lemur do you have? lemp11?

jacobgkau:
@n3m0-22 The style of those brackets/parenthesis should not match other entries because this is not part of a top-level bullet point like the rest are.

jacobgkau:
After seeing how it looks, I guess that's fine. This article is getting unwieldy in general.

jacobgkau (approved):

n3m0-22 (approved):

jlangson:
@jacobgkau It's a lemur 11

PR #1066, opened 2022-09-23
Resolving ProtonVPN Killswitch Issues, Install Pop updates

Ampersandstorm:
I created this article after battling the unexpected nuances that can occur when enabling ProtonVPN's permanent kill switch feature. 

install-pop.md updates:

- updated all images for 22.04
- added steps for selecting keyboard options during install
- cleaned up extraneous wording and punctuation issues
- fixed headers not in title case
- placed system requires into table
- added section for selecting device at boot
- add section about secure boot
- add note about changing boot order or removing bootable disk after installation and reboot
- moved unused images to folder


content/fix-pvpn-killswitch.md
```diff
@@ -0,0 +1,61 @@
...
+
+facebookImage: /_social/article
+twitterImage: /_social/article
+
+hidden: false
+section: software
+tableOfContents: true
+---
+
+ProtonVPN's kill switch will disable internet connectivity when you are not connected to a VPN server. This ensures the true IP address if you computer is never leaked to the internet. If the kill switch is set to `Permanent`, then your computer will be unable to connect to the internet, even if ProtonVPN isn't running (i.e., at system start up), the GUI application becomes inaccessible, or if ProtonVPN was improperly removed from the system.
```
leviport, line 22: >the true IP address if you computer is never leaked

should be 

>the true IP address of your computer is never leaked

content/install-pop.md
```diff
@@ -22,68 +22,105 @@ tableOfContents: true
...
+## Requirements
 
-## Download the Pop!\_OS.iso image
+### System Specifications
 
-From <https://pop.system76.com/>, choose **DOWNLOAD**, and then choose from the current release (default), or the Long Term Support (LTS) image. The download dialog includes the current _Requirements_, and download _Filesize_, as well as the image checksums used to verify the file once downloaded.
+| Component | Requirement | Recommended |
+|-----------|-------------|-------------|
+| CPU       | 64-bit x86 architecture |
+| RAM       | 2 GB        | 4 GB        |
```
leviport, line 32: I wonder if it's a good time to bump the minimum RAM requirement up a bit. Running Pop on 2gb is pretty abysmal, especially if it's installed on slow storage devices since it has to rely on swap so much.

leviport: Oh also, the CPU requirement only applies to normal Pop, since now we have the Pi image...

content/install-pop.md
```diff
@@ -22,68 +22,105 @@ tableOfContents: true
...
 
-## Make Bootable Drive
+### Installation Media
 
-Please see our instructions for making and booting from a live disk of Pop!\_OS [here](/articles/live-disk/).
+The Pop!\_OS.iso can be easily flashed to a USB drive. See our [Live Disk Creation](/articles/live-disk/) article for instructions to create a bootable USB drive in Windows, macOS, and Linux.
+
+### Secure Boot
+
+Systems with Secure Boot enabled must disable this feature before installing Pop!\_OS. Secure boot can be disabled in the BIOS of most computers; however, the process to disable secure boot will vary by laptop and motherboard model.
```
leviport, line 41: >Systems with Secure Boot enabled must disable this feature

This sentence feels a little awkward to me. Maybe it could be streamlined to something like:

>Secure boot must be disabled before installing Pop!\_OS

content/install-pop.md
```diff
@@ -22,68 +22,105 @@ tableOfContents: true
...
+| Laptop - Open Firmware | ESC      | ESC (select one time boot option) |
+| Laptop - Proprietary   | F2       | F7                               |
+| Older Laptops          | Depends on the system | F1                  |
+| Thelio                 | Del      |  F8 or F12                       |
+| Meerkat                | F2       | F10                              |
 
 ## Installing Pop!\_OS
 
-The first step after booting from your new live disk is to select your language:
+Once your bootable USB drive is selected as a boot device, your computer will into a live Pop!\_OS environment.
```
leviport, line 57: >your computer will into a live

Missing a word

content/install-pop.md
Ampersandstorm: I made a separate PR for install-pop.md updates, but I guess those changes are a part of this PR too. I'm not sure how this happened.

leviport: Oh I see. Maybe you didn't `git checkout master` before creating this new branch?

Ampersandstorm: I think I forgot to close vscode and just kept working after creating the new branch.

Ampersandstorm: I updated requirements to 4 gb min, 8 gb recommended, and added ARM (RAS Pi 4) as an architecture.

content/install-pop.md
Ampersandstorm: Sounds good. I assumed there was some significant number of machines that did not have a secure boot.

content/install-pop.md
ahoneybun: Do we want to change ARM to ARM64 as there is no image for 32-bit ARM?

ahoneybun (approved):
Other then my comment about changing to ARM to ARM64 this looks great!

content/install-pop.md
leviport: It says RASPI 4, so I think that's probably specific enough.

leviport (approved):

PR #1069, opened 2022-10-10
Clarify hibernation risks on Pop!_OS 22.04

13r0ck:


content/enable-hibernation.md
```diff
@@ -20,6 +20,10 @@ section: software
 tableOfContents: true
 ---
 
+## Disclaimer
+
+Hibernation is not officially supported on Pop!_OS as of 22.04. This documentation is supplied for your own explorative purposes. Any changes to your system configuration may break hibernation, or hibernation may not work at all. As well, any support or assitance may not be provided.
```
jacobgkau, line 25: > As well, any support or assitance may not be provided.

`assitance` is a typo.

Can the sentence be simplified to just `Support or assistance may not be provided.`?

13r0ck: Thanks! Fixed

jacobgkau (approved):
Seeing that hibernation is not currently a tested/default feature, I don't see any problems with this wording from a QA standpoint, as long as it's representative of both engineering's and support's stances on the matter.

13r0ck:
@system76/support

garnerlogan65 (approved):
Maybe add "Hibernation is outside of the official scope of System76 support" before "Support or assistance may not be provided"?

jacobgkau:
I think that would be a little redundant with the first and last sentences.

PR #1071, opened 2022-10-13
fix upgrade-older-releases link

n3m0-22:
This is to fix the issue mentioned on Mattermost https://chat.pop-os.org/pop-os/pl/mhz6sod7sjrf7gsgup1zuiojxo


leviport:
Lol, beat me by a hair: https://github.com/system76/docs/pull/1072

leviport (dismissed):

jacobgkau:
Would it make more sense to remove `(pre 20.04)` from the heading since the section also addresses 20.10 (which is not pre-20.04)? Or should 20.10 be moved out of the section?

n3m0-22:
I think it makes sense to remove `(pre 20.04)` and leave the `21.10` instructions as there are still users asking about upgrading from it. Where would it be moved to if not in this section, or would you remove it completely?

jacobgkau:
> Where would it be moved to if not in this section, or would you remove it completely

I would probably move it into its own section. Anything before 20.04 should go to 20.04, then to 22.04. Anything after 20.04 should go straight to 22.04.

I wouldn't be opposed to just reverting the `(pre 20.04)` from the heading for now, and then the section can have sub-sections for pre-20.04 and 20.10-21.10 later.

n3m0-22:
How about something like this:

Screenshot from 2022-10-13 20-17-30

jacobgkau:
> How about something like this:

The leftmost column is redundant with the middle column. They both list versions.

The advise to refresh from recovery/USB for 20.10-21.10 isn't really in line with the advise being given for pre-20.04. Any version can be upgraded manually via apt, and probably any version can be refresh installed to the latest version. That's more a matter of if the user wants to spend time fixing it the "correct" way or just get their system working as fast/simple as possible.

With that in mind, I would imagine the post-20.04 section would still have the sed and apt commands to upgrade manually, just switching things to `jammy` instead of `focal`. But we should keep the refresh call-out as an alternative for any version it will work for. (The refresh call-out should also not make it sound like it would only work from recovery; refresh from live USB is also an option.)

jacobgkau (approved):
The current changeset fixes the broken link and the inaccurate header.

PR #1078, opened 2022-11-03
Adds missing models and removes system76-tested dock

ahoneybun:


leviport:
Why is the Plugable dock being removed?

thomas-zimmerman:
Customer reports that "the System76 tested" dock is not working on modern kernel releases. This dock was last tested on 20.04

PR #1085, opened 2022-11-21
Fix a minor confusion

realKarthikNair:
````nvme0n1p3```` won't necessarily be the Pop!_OS partition


jacobgkau:
Fixed the grammar, duplicated the new line everywhere it's relevant (excluded the LUKS section since our installer only offers LUKS with the default partitioning scheme), and made the note more general since there are other situations where it might apply.

jacobgkau (approved):

thomas-zimmerman (approved):
Good reminder that partitions will change if you do something different to start off an install.

realKarthikNair:
Thank you so much guys for fixing that quickly

PR #1086, opened 2022-11-23
Add gaze17, oryp9, and oryp 10

zachary-kaelan:
Bringing it in line with what is listed here: https://github.com/pop-os/gnome-shell-extension-system76-power/blob/master/src/extension.ts


content/graphics-switch-pop.md
```diff
@@ -29,9 +29,9 @@ The following System76 laptops have these switchable graphics capabilities:
 
 - Adder WS (addw1, addw2)
 - Galago Pro (galp5 - NVIDIA models only)
-- Gazelle (gaze14, gaze15, gaze16)
+- Gazelle (gaze14, gaze15, gaze16, gaze17-3050, gaze17-3060-b)
```
leviport, line 32: We could probably simplify to just `gaze17` here, since `gaze16` doesn't specifically call out each sub-variant.

n3m0-22: That's how it was originally. I was trying to match the file referenced as well as the `oryp4` models. Since `gaze16` has been simplified, I'll change it back. Would it make sense to simplify the `oryp4` and `oryp4-b` as well?

n3m0-22 (approved):
Thank you for your contribution.

ahoneybun (approved):

ahoneybun:
Before we merge this we need to update the Ubuntu article as well.

jacobgkau:
> Before we merge this we need to update the Ubuntu article as well.

Done.

PR #1087, opened 2022-12-01
Remove Docking Station article

ahoneybun:
This removes this article, this was approved by Thomas and Emma on the Careteam.


leviport (approved):
I still think it would be useful to revamp the article rather than remove it. Instead of mentioning any specific models, it could just be a brief overview of DP MST, PD over USB-C, Thunderbolt vs. native vs. DisplayLink, etc. 

But if you want it gone, then it can be gone.

ahoneybun:
> I still think it would be useful to revamp the article rather than remove it. Instead of mentioning any specific models, it could just be a brief overview of DP MST, PD over USB-C, Thunderbolt vs. native vs. DisplayLink, etc.
> 
> But if you want it gone, then it can be gone.

It's best to remove it as neither QA or Support have the time to update this article or perform tests on the system76-tested docking station. We'd rather remove incorrect information in this case.

jacobgkau:
We've had a lot of community members make contributions to this article to report their own experiences with various docks. Do we have somewhere else that we can point people to for that kind of thing? I know support articles aren't a wiki, but there is also no official Pop!_OS or System76 wiki (or unofficial that I'm aware of, since Pop!_Planet is defunct.)

PR #1089, opened 2022-12-11
Add an alternative for fixing audio

pavelexpertov:
After facing an issue of inability to use current instructions to fix audio issues on Pop Os 22.04 version, I stumbled upon this [link](https://github.com/pipewire-debian/pipewire-debian/issues/49#issuecomment-1322671652) where one can just reinstall packages and restart the services. 

The instructions should be included to provide alternative instructions as to fix the issue. Of course, provide feedback whether this can be improved upon.

### DIFF
--- a/content/audio.md
+++ b/content/audio.md
@@ -32,6 +32,12 @@ If the system is not playing audio, first try restarting the audio daemon:
 
 This set of commands first restarts the sound daemon and removes the user's configuration for PulseAudio. On systems still using PulseAudio as a server, it restarts the PulseAudio server, which will create new default audio configuration files.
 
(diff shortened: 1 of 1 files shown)


jacobgkau:
Reinstalling packages (including these two packages) is already covered lower in the article, and restarting the services is already covered in the table just above the addition.

pavelexpertov:
Cool, just re-read the docs and found the commands that you mentioned. It's just a shame that there's no short indication to say. "If that doesn't work, explore the other commands section". Do you think that will suffice for reading improvement? Otherwise, the PR can be closed. @jacobgkau

PR #1094, opened 2022-12-28
Update Ubuntu-upgrade

thomas-zimmerman:
Customer feedback on the upgrade article. Expand on PPA changes. Add some additional search keywords to find this article.


jacobgkau:
As of right now, the only thing this PR is doing is replacing one image with another. Was that commit supposed to include the other changes that its description listed?

thomas-zimmerman:
I'll need to revisit as the text changes I made didn't seem to land with this PR

PR #1096, opened 2023-01-06
Update booloader.md for encrypted UEFI systems on Ubuntu

adfernandes:
Update booloader.md for encrypted UEFI systems on Ubuntu


ahoneybun:
@adfernandes thank you for the PR! I think this article needs a fairly large rewrite or maybe reorganization to be more clear and easier to understand. I'll need to do that from another PR though as I don't believe I can edit your branch.

PR #1097, opened 2023-01-08
Fix link that is currently broken

emujoyce:
The existing link is broken. Removing the ".md" should resolve the issue and link properly.


leviport (approved):
Looks like that fixes the link. Thanks for the contribution!

thomas-zimmerman (approved):
Let's get this fixed! Thank you for the PR.

PR #1102, opened 2023-01-29
add Krita in media production on pop

realKarthikNair:


content/media-production-on-pop.md
```diff
@@ -86,3 +87,9 @@ Install Directly From the Pop!\_Shop
 Flowblade is a multitrack non-linear video editor with a set of industry standard editing tools suitable for beginners and masters alike to help make your vision a reality of image and sound.
 
 Install Directly From the Pop!\_Shop
+
+### Krita
+
+Krita is a professional FREE and open source painting program. It is made by artists that want to see affordable art tools for everyone.
```
leviport, line 93: The word "free" should not be in all caps, to keep it consistent with the rest of the article. The other application descriptions also don't talk about the motivations of the developers, so the second sentence shouldn't be there. How about something like this: 

>Krita is a free and open source raster graphics editor that is primarily focused on digital drawing and painting. It features an OpenGL-accelerated canvas, non-destructive layers and masks, an advanced brush engine, and support for many drawing tablets.

leviport:
Thanks for the contribution! Since GIMP and Inkscape are both on the list, I think including Krita makes a lot of sense.

PR #1105, opened 2023-02-02
Verify UEFI boot: Expected output correctly placed

mabhijithn:
UEFI boot verify command's expected output was put in the subsequent section incorrectly. This has been changed to its correct section.


jacobgkau (approved):
This article's flow can probably be simplified further, but the current order doesn't make sense and this fix is good.

thomas-zimmerman (approved):
Good change. Thank you for the PR!

mabhijithn:
My pleasure! Noticed it when actually using the guide.. :-)

PR #1107, opened 2023-02-03
fix: paru sync command

### LINKED ISSUE #1106: Invalid `paru` command for System76 Driver and Software on Arch
The following installation guides use `paru -s` instead of `paru -S`, which seems to be a typo:

```
       -s, --ssh
              Clone the AUR package using SSH (e.g.: a read-write remote).
```

Docs using `paru -S`:

https://github.com/system76/docs/blob/d133c79ec3f66a6d0860f49251c78d9b2b77da8f/content/system76-software.md?plain=1#L37

Docs using `paru -s`:

https://github.com/system76/docs/blob/d133c79ec3f66a6d0860f49251c78d9b2b77da8f/content/system76-driver.md?plain=1#L137-L147

https://github.com/system76/docs/blob/d133c79ec3f66a6d0860f49251c78d9b2b77da8f/content/system76-software.md

I would be happy to open a PR to fix this!

qin-guan:
Closes #1106

### DIFF
--- a/content/system76-driver.md
+++ b/content/system76-driver.md
@@ -134,17 +134,17 @@ makepkg -si
 **Note:** By default, <u>Paru</u> uses VIM keyboard shortcuts, so when you see a `:`, press the <kbd>q</kbd> key to continue. You may also need to confirm the import of some public keys using the <kbd>y</kbd> key.
 
 ```bash
-paru -s system76-firmware-daemon
+paru -S system76-firmware-daemon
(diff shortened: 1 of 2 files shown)


leviport (approved):
I haven't used `paru` before, but according to the [README](https://github.com/Morganamilo/paru), it looks like `-S` is indeed the correct flag for installing things. Thanks for the fix!

ahoneybun (approved):
LGTM :tada:

PR #1108, opened 2023-02-05
Uppercase Hash Algorithm for CertUtil

acropup:
CertUtil in Windows 7 requires the HashAlgorithm option to be uppercase. In this case, SHA256 needs to be uppercase. In Windows 10, case does not matter. I have not verified with other versions, but I presume that all versions of windows will work with SHA256 specified in uppercase.

### DIFF
--- a/content/live-disk.md
+++ b/content/live-disk.md
@@ -69,13 +69,13 @@ Assuming you downloaded Pop!_OS to your Downloads folder, open the Command Promp
 #### For Intel/AMD iso
 
 ```
-CertUtil -hashfile Downloads\pop-os_22.04_amd64_intel_4.iso sha256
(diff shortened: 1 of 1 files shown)


leviport:
Is it worth making an update for an OS that reached EOL in 2020?

jacobgkau (approved):
Even on Windows 10, the CertUtil output suggests uppercase is the proper syntax, even if it's case-insensitive:

certutil

I think this is similar to how the utility is technically `CertUtil`, although you can type `certutil` or `CeRtUtIl` because the Windows CLI is case-insensitive in many situations. The actual capitalization is still more correct.

This is a good change.

thomas-zimmerman (approved):
Good change; getting this correct on where you're running the commands is helpful.

leviport:
Thanks for the fix!

PR #1111, opened 2023-02-17
Add Krita to media production list

leviport:
I'm not sure why https://github.com/system76/docs/pull/1102 got closed, but I still think it's a good addition. I've recreated it here, with a tweaked description for Krita. 

I also added a few keywords for apps that are listed, but didn't have keywords yet.


n3m0-22 (approved):
I think this is a good addition as well. This looks good to me.

thomas-zimmerman (approved):
Looks good

PR #1114, opened 2023-03-09
Update touchpad disable

ahoneybun:


content/touchpad.md
```diff
@@ -20,9 +20,15 @@ tableOfContents: true
 
 Press <kbd>Fn</kbd>+<kbd>F1</kbd> to turn your laptop touchpad on/off.
 
+>**Note**: This shortcut is not available on the Pangolin (pang12).
```
jacobgkau, line 23: The Pangolin does have a touchpad disable shortcut, it's just Fn-F6 instead of Fn-F1: https://tech-docs.system76.com/models/pang12/external-overview.html

ahoneybun: Should be fixed in the latest commit!

content/touchpad.md
```diff
@@ -20,9 +20,15 @@ tableOfContents: true
...
 Press <kbd>Fn</kbd>+<kbd>F1</kbd> to turn your laptop touchpad on/off.
 
+>**Note**: This shortcut is <kbd>Fn</kbd>+<kbd>F6</kbd> on the Pangolin (pang12).
+
 To configure your touchpad, press the Super Key <kbd><font-awesome-icon :icon="['fab', 'ubuntu']"></font-awesome-icon></kbd>, <kbd><font-awesome-icon :icon="['fab', 'pop-os']"></font-awesome-icon></kbd> then type <u>Mouse & Touchpad</u> and click on the result. The touchpad settings are under the **Touchpad** tab.
 
-![Mouse Search](/images/touchpad/mouse.png)
+![Mouse Search](/images/touchpad/touchpad-mouse.png)
+
+From here the touchpad can be disabled only while typing like in the screenshot below:
```
jacobgkau, line 29: > From here the touchpad can be disabled only while typing like in the screenshot below:

A comma should go after `From here`, and also before `like`.

content/touchpad.md
```diff
@@ -20,9 +20,15 @@ tableOfContents: true
...
+
+![Mouse & Touchpad Settings](/images/touchpad/touchpad-mouse-settings.png)
 
 <u>GNOME Tweaks</u> has a few other options for the touchpad like disabling it while typing, the acceleration profile and mouse click emulation.
```
jacobgkau, line 33: Part of this sentence is now redundant since disable-while-typing is now mentioned and shown above. (The new screenshot is also redundant with the Tweaks screenshot below this sentence.)

ahoneybun: Should be fixed in latest commit.

content/touchpad.md
ahoneybun: Should be fixed in latest commit.

content/touchpad.md
jacobgkau: Sorry, the sentence did need to be modified, but the screenshot itself wasn't redundant; I didn't realize the new one was of Settings and not GNOME Tweaks since the header bar wasn't included in the new screenshot.

I'm going to change the new Settings one so it includes the window border to better match the Tweaks screenshot, and to highlight the setting being referred to.

jacobgkau: Hmm, we've actually moved emulation into the main Settings, so that also needs to be removed from the Tweaks callout.

jacobgkau: Also, I'm not sure why we're explaining how to enable this setting when the setting is already enabled by default.

jacobgkau: Made some adjustments in https://github.com/system76/docs/pull/1114/commits/ad3a7c73576be9edef29eda2f46d9d642414c9f8 & https://github.com/system76/docs/pull/1114/commits/0abf88634cea88c4acd9b1005cd1ce2379a47350.

jacobgkau (approved):

thomas-zimmerman (approved):
Let's get this published!

PR #1115, opened 2023-03-18
Update boot-menu.md

edsix6:
shouldn't it be "When you see the System76 logo on the screen, press and hold the appropriate key for your system:"  Many other pages are similarly wrong.

### DIFF
--- a/content/boot-menu.md
+++ b/content/boot-menu.md
@@ -16,7 +16,7 @@ section: software
 tableOfContents: true
 ---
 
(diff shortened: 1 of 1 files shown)


leviport:
Not exactly. Many of the older System76 laptops with proprietary firmware don't display a System76 logo. And some, like the pang12 leave an extremely narrow window, so if you wait for the splash screen, you might miss it. Hitting BIOS keys right from when the machine is turned on is definitely more likely to get the machine to go into BIOS setup on the first try.

PR #1117, opened 2023-03-20
boot-menu: Slightly reformat & add pang12

### LINKED ISSUE #1116: BIOS key for pang12 needs to be added to Access BIOS article
https://support.system76.com/articles/boot-menu

This one might be a little tricky since it's `Esc` for BIOS, and the boot menu is also in BIOS, meaning it's almost the same as open firmware. We might have to call it out specifically by model, along with any future models in this line that share the same BIOS.

jacobgkau:
Closes #1116.

image

### DIFF
--- a/content/boot-menu.md
+++ b/content/boot-menu.md
@@ -20,7 +20,7 @@ The BIOS or boot menu can be accessed by holding specific keys down during boot.
 
 | Model                                                       | BIOS key | Boot Menu key                     |
 |:-----------------------------------------------------------:|:--------:|:---------------------------------:|
-| Laptops:<br/>All Open Firmware models<br/>Pangolin (pang12) | ESC      | ESC:<br/>Select one-time boot (Open Firmware)<br/>or Save & Exit -> Boot Override. |
+| Laptops:<br/>All Open Firmware models<br/>Pangolin (pang12) | ESC      | ESC:<br/>Select `One Time Boot` (Open Firmware)<br/>or `Save & Exit` → `Boot Override`. |
 | Laptops:<br/>Most proprietary firmware models               | F2       | F7                                |
 | Older laptops                                               | Depends on the system | F1                   |
 | Thelio desktops                                             | Del      | F8 or F12                         |
--- a/content/boot-menu.md
+++ b/content/boot-menu.md
@@ -16,15 +16,15 @@ section: software
 tableOfContents: true
 ---
 
-To enter BIOS for most System76 laptops, hold the <kbd>F2</kbd> key immediately after turning your computer on.  For the boot menu, hold down the <kbd>F7</kbd> key.  For older laptops, hold the <kbd>F1</kbd> key.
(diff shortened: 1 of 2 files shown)


content/boot-menu.md
```diff
@@ -16,15 +16,15 @@ section: software
...
-| Laptop - Open Firmware | ESC      | ESC(select one time boot option) |
-| Laptop - Proprietary   | F2       | F7                               |
-| Older Laptops          | Depends on the system | F1                  |
-| Thelio                 | Del      |  F8 or F12                       |
-| Meerkat                | F2       | F10                              |
+The BIOS or boot menu can be accessed by holding specific keys down during boot. Refer to the table below for the keys on each model:
+
+| Model                                                       | BIOS key | Boot Menu key                     |
+|:-----------------------------------------------------------:|:--------:|:---------------------------------:|
+| Laptops:<br/>All Open Firmware models<br/>Pangolin (pang12) | ESC      | ESC (select one-time boot option) |
```
leviport, line 23: I wonder if the `(select one-time boot option)` note should be tweaked to account for pang12's boot menu:

IMG_20230320_155402_417

I'm thinking maybe something like, "(select one-time boot option in open firmware, or use the boot override menu in "save & exit" on pang12)," but it's awfully wordy.

jacobgkau:
Is this better?

image

content/boot-menu.md
jacobgkau: Is this better?

image

jacobgkau: I guess we might want the page that it's on for proprietary.

leviport:
Yeah that looks good! I think the case should be consistent though. Either make (Open Firmware) lowercase or capitalize (proprietary firmware).

content/boot-menu.md
jacobgkau: Here's another option:

image

Is that clear enough?

jacobgkau:
> Yeah that looks good! I think the case should be consistent though. Either make (Open Firmware) lowercase or capitalize (proprietary firmware).

Open Firmware is a product name, proprietary firmware is a description.

leviport:
Ah, fair, I hadn't thought of it as a product name.

leviport (approved):

jacobgkau:
I will need to either allow the `<br>` tag in the Markdown linter or find another way to break the line within the table cells.

jacobgkau:
I allowed this in the linter because Markdown doesn't seem to have a way to include line breaks within table cells by default (some specific flavors of Markdown have their own ways of doing this, but I didn't quickly find one that works with the Nuxt parser that we're using.)

jacobgkau:
This is what it looks like after the last commit:

image

My first instinct was that the code tags made it look to busy, but it looks okay to me after considering it further. I can revert the last commit if other people don't like it.

ahoneybun (approved):
LGTM

PR #1121, opened 2023-04-11
Updates to pop-recovery

thomas-zimmerman:
Customer feedback -- found that we were not clear that once you accessed drives, installer was not working for him with errors that are not explained/clear

### DIFF
--- a/content/pop-recovery.md
+++ b/content/pop-recovery.md
@@ -57,13 +57,13 @@ If the `Refresh Install` option is not present on the install screen, one of two
 
 Once Recovery has booted, the <u>Pop Installer</u> will start automatically.  If the system needs to be reinstalled, go ahead and continue the installation steps as demonstrated [here](/articles/install-pop/).
 
-If the existing install is encrypted, please see the [encrypted disk](#encrypted-disk) instructions.
+If the existing install is encrypted, please see the [encrypted disk](#encrypted-disk) instructions. 
 
 ## Repair
 
 If the existing OS install needs to be repaired, the installer application should be closed. An app menu is located in the top-left of the screen with the name of the currently running application (in this case: "Install Pop!\_OS"). Click on the app menu and select `Quit`. Alternatively, you can use the installer app to select keyboard and language settings, and then click the `Try Demo Mode` button in the lower-left corner of the install page.
 
->**Note:** Do not choose any install or repair options, as this could result in data loss.
+>**Note:** Do not choose any install or repair options, as this could result in data loss. Also, once you have setup or accessed any of the filesystems, you will need to boot to the recovery environment anew to use any of the install options.
 
 Follow the instructions below to access the existing OS drive:
 


ahoneybun (approved):
Changes look good and make sense.

garrettjwilke (dismissed):

content/pop-recovery.md
```diff
@@ -63,7 +63,7 @@ If the existing install is encrypted, please see the [encrypted disk](#encrypted
 
 If the existing OS install needs to be repaired, the installer application should be closed. An app menu is located in the top-left of the screen with the name of the currently running application (in this case: "Install Pop!\_OS"). Click on the app menu and select `Quit`. Alternatively, you can use the installer app to select keyboard and language settings, and then click the `Try Demo Mode` button in the lower-left corner of the install page.
 
->**Note:** Do not choose any install or repair options, as this could result in data loss.
+>**Note:** Do not choose any install or repair options, as this could result in data loss. Also, once you have setup or accessed any of the filesystems, you will need to boot to the recovery environment anew to use any of the install options.
```
leviport, line 66: I feel like these should be two different notes. Tacking that onto a note that warns people _not_ to do an install seems like it just doesn't fit.

content/pop-recovery.md
```diff
@@ -63,7 +63,9 @@ If the existing install is encrypted, please see the [encrypted disk](#encrypted
 
 If the existing OS install needs to be repaired, the installer application should be closed. An app menu is located in the top-left of the screen with the name of the currently running application (in this case: "Install Pop!\_OS"). Click on the app menu and select `Quit`. Alternatively, you can use the installer app to select keyboard and language settings, and then click the `Try Demo Mode` button in the lower-left corner of the install page.
 
->**Note:** Do not choose any install or repair options, as this could result in data loss.
+>**Note:** Do not choose any install or repair options, as this could result in data loss. 
+
+>**Note:** Also, once you have setup or accessed any of the filesystems, you will need to boot to the recovery environment anew to use any of the install options.
```
leviport, line 68: How about, `**Note:** If you have accessed any filesystems and intend to install or refresh afterwards, the machine should be rebooted into a fresh recovery environment to clear any created mount points.` instead? Still feels a bit clumsy to me, but it feels better to lead with the situation in which the note is important for the reader.

thomas-zimmerman: I like this update better. Looking at the Live version of this change looks good to me.

content/pop-recovery.md
```diff
@@ -63,9 +63,7 @@ If the existing install is encrypted, please see the [encrypted disk](#encrypted
 
 If the existing OS install needs to be repaired, the installer application should be closed. An app menu is located in the top-left of the screen with the name of the currently running application (in this case: "Install Pop!\_OS"). Click on the app menu and select `Quit`. Alternatively, you can use the installer app to select keyboard and language settings, and then click the `Try Demo Mode` button in the lower-left corner of the install page.
 
->**Note:** Do not choose any install or repair options, as this could result in data loss.
-
->**Note:** Also, once you have setup or accessed any of the filesystems, you will need to boot to the recovery environment anew to use any of the install options.
+>**Note:** Do not choose any install or repair options, as this could result in data loss. If you have accessed any filesystems and intend to install or refresh afterwards, the machine should be rebooted into a fresh recovery environment to clear any created mount points.
```
leviport, line 66: Now it's back to being a single comment. I still think these are two separate points, each deserving of their own note.

leviport (changes requested):

content/pop-recovery.md
ahoneybun: It breaks the CI from what I can tell but I could be wrong.

jacobgkau: If wanted formatting breaks the CI, then the CI rules should be adjusted.

That said, if we have two separate things to say, then maybe a regular paragraph or a bullet-point list would make more sense than two highlighted notes in a row (which dilutes the highlighting, anyway.)

jacobgkau: It looks like the Markdown linter is interpreting the two blockquotes as one large blockquote, so adding an additional `>` on the blank line between them might help?

jacobgkau:
The paragraph preceding these notes still referenced a pre-COSMIC app menu as the primary way to quit the installer. That probably also contributed to confusion, and I've removed it now.

jacobgkau:
image

This is what it looks like right now.

jacobgkau:
Changing the first `Note:` to `Caution:` makes it looks a little less messy.

image

jacobgkau:
image

Added a dock method back to replace the app menu method that I removed.

jacobgkau (approved):
Please squash this when merging.

ahoneybun (approved):
LGTM

leviport (approved):

PR #1122, opened 2023-04-13
[FIX] kernelstub: wrong command for adding kernel options

pparidans:
The `-o` option actually replaces the whole boot options' list passed to the kernel instead of adding to the configuration (if they aren't present already), which is done using `-a` (or the lengthier `--add-options`).


leviport (approved):
Thanks for the correction!

ahoneybun (approved):

PR #1133, opened 2023-05-22
Update wireless.md

matt384:
Added section for dual boot systems. I had an issue with Wifi card not found after booting into windows 11 and then back to pop_os. I spent a few hours to find this to be the issue. I have confirmed on my machine disabling fast boot makes the wifi card discoverable by pop_os again.


ahoneybun:
Thank you for the PR! Could you reword it so that is not a directly copy from that page by chance?

PR #1134, opened 2023-05-23
Add bullet point to Before You Begin section in system-firmware doc

nonergodic:
Tried running the firmware update process on my Thelio mega r2 and ran into the issue that my monitor remained black. Eventually realized that the problem was that my screen was only coming on after the x server had started, i.e. the bios apparently doesn't support the display port ports of the GPU but only the HDMI port.

Adding a note to the documentation to hopefully spare others a similar fate (especially when not familiar with their new machine yet which increases general uncertainty).


nonergodic:
I have two RTX4090s in my rig and get no signal on boot on either of them until the login screen. :man_shrugging:

jacobgkau:
I don't have permissions in @nonergodic's fork to force-push and improve the last commit message that GitHub generated.

@leviport I've reworded this so it's less specific and also covers laptops. Do you want to merge it or close it? Up to you.

I don't think it hurts to mention, but also don't think it's necessary to actually reboot and check before applying the update (this is more in line with something I'd ask during troubleshooting if someone says it's not working, but this article doesn't currently have a "troubleshooting" section, and it's more straightforward to mention it upfront).

leviport:
Ah, I'm glad this came back up again, because I thought of something new. I think it is possible that the VGA switch could be turned on on your motherboard (see label 27 on page 1-2 in the [motherboard manual](https://dlcdnets.asus.com/pub/ASUS/mb/SocketTRX4/Pro_WS_WRX80E-SAGE_SE_WIFI/E19401_Pro_WS_WRX80E-SAGE_SE_WIFI_UM_V2_WEB.pdf?model=Pro%20WS%20WRX80E-SAGE%20SE%20WIFI0). If that gets turned on, I've seen mega-r2 prioritize its built-in VGA output over discrete graphics cards. If you see an extra low resolution display (something like 640x480) in display settings.

leviport (approved):
Approving because I don't think this note hurts anything to include either way.

ahoneybun (approved):

PR #1141, opened 2023-06-05
Install libdvd-pkg before running dpkg-reconfigure

mrvictory1:
Installing ubuntu-restricted-extras does not install libdvd-pkg and the dpkg-reconfigure command mentioned in the guide fails because libdvd-pkg is not installed.


jacobgkau (changes requested):
Confirmed `libdvd-pkg` is not included with `ubuntu-restricted-extras`. However, it was included in the "manual"/"minimal" alternative commands already provided. It should probably either be removed from those two "manual" lines if we explicitly state to install it later, or it should just be added to the first line alongside `ubuntu-restricted-extras`.

PR #1143, opened 2023-06-13
Windows dual boot info effecting wifi card access

jonknapp:
Add info about disabling Windows "Fast Startup" which was the cause of my missing wifi card after a driver update from the Windows side.


leviport (approved):
I'm still not sure how prevalent this issue is, but I definitely see it being helpful if someone happens to encounter it. I'm alright with the addition. Thanks!

ahoneybun (approved):

PR #1145, opened 2023-06-20
Add section for tech docs and guides

ahoneybun:
Address feedback from a support ticket and helps promote tech-docs.

Change:

Screenshot from 2023-06-20 09-11-20


thomas-zimmerman (approved):
LGTM

leviport (changes requested):
I do have to agree with @Raikiri about the location of this being rather awkward since it's pretty unrelated to contacting us. I think it would be best to add a section in the "Before you open a ticket" article. The left column would _kinda_ work, but it still feels a bit awkward and out of place.

What do you think of adding a section to the "Before you open a ticket" article?

Raikiri:
Putting it into "Before you open a ticket" is basically putting into the last possible place somebody might want to check right before creating a ticket that says "I can't find your technical documentation". So no, I don't think it's a good place, I think the link needs to be in the topmost location in the navigation hierarchy, maybe even put it into the header:
image


Honestly "Support articles" seems like a really logical place for the list of support articles to be.

ahoneybun:
@Raikiri that location would not be part of the docs project and instead with the web team so it's not an option at least from this PR.

leviport:
>Putting it into "Before you open a ticket" is basically putting into the last possible place somebody might want to check right before creating a ticket that says "I can't find your technical documentation". 

I meant that question for Aaron, but thanks for the input. My impression is that's pretty much the exact idea we were going for. This is a link to a support article that has various links to tech-docs for each model. I believe having that support article amongst the rest is still sufficient, but I agree with Aaron that it can probably be improved. Adding a link to a support article into the header for the whole site, next to a link to the list of support articles, is not something I'm super keen on. And we certainly wouldn't be able to get Marketing, Web, Sales, etc. on board with such a change either.

Raikiri:
Another thing I wanted to mention: if it's going to be named "Find Documentation For Your Model" among a million other articles, I expect it to be very hard to find: maybe it's a me problem, but when I want to repair my laptop, I never search for "Documentation", I search for "repair guide", "<model name> internal view", "disassembly", etc. "Documentation" strikes me as data sheets rather than a guide.

Pardon me if I sound too nitpicky and I don't really have a solution to this problem. I'm just providing you a data point of my own experience: since from my past experience I know that you guys have this documentation, after some struggle I do manage to find it every time I need it. But if I was on your page for the first time, a good chance I wouldn't be able to.

leviport:
>I search for "repair guide", " internal view", "disassembly", etc. 

That's a good point, and I think it'd be good to add those to the search terms for that article. "Components" and "guide" both bring it up, but adding more terms is easy enough.

ahoneybun:
@leviport @Raikiri I've made some changes so if you want to give this another review let me know.

content/before-you-open-a-support-ticket.md
```diff
@@ -11,6 +11,9 @@ keywords:
  - self-help
  - tech-support
  - troubleshooting
+ - repair guide
+ - internal view
+ - disassembly
```
leviport, line 16: Should the keywords be in this article as well? It seems to me that it'd be better if they were only in the guides article

leviport (approved):

leviport:
(I just removed a stray space that was left over from where those keywords were deleted)

Raikiri:
I still can't find the documentation while searching for "repair guide" or "internal view". Even if I do stumble upon "Documentation for your model", it's not immediately obvious that it's what I'm looking for. But it's obviously better than none at all.

ahoneybun:
> I still can't find the documentation while searching for "repair guide" or "internal view". Even if I do stumble upon "Documentation for your model", it's not immediately obvious that it's what I'm looking for. But it's obviously better than none at all.

The keywords do not effect the search, it seems to be based on title and description.

thomas-zimmerman (approved):
Let's get this change live and see if this helps. Changes look fine on the live view.

PR #1148, opened 2023-06-26
Add missing `addw3`, `gaze18`, `oryp11` models

ahoneybun:
This adds new models to the list with switchable graphics. I wonder if we should just put the models that we started offering switchable graphics since every system in the 13th Gen Intel has a dGPU so it has switchable graphics. Then we would not have to update this page as often.

### DIFF
--- a/content/graphics-switch-ubuntu.md
+++ b/content/graphics-switch-ubuntu.md
@@ -27,7 +27,7 @@ tableOfContents: true
 
 Many modern laptops with NVIDIA graphics cards have switchable graphics, which allows users to switch their primary GPU between the CPU's integrated graphics processor and the dedicated NVIDIA graphics card. The following System76 laptops have these switchable graphics capabilities:
 
-- Adder WS (addw1, addw2)
+- Adder WS (addw1, addw2, addw3)
 - Bonobo WS (bonw15)
 - Galago Pro (galp5 - NVIDIA models only)
 - Gazelle (gaze14, gaze15, gaze16, gaze17, gaze18)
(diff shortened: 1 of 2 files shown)


leviport (changes requested):
Missed `addw3` and `oryp11`

If there's an elegant way to do an "all 13th gen Intel + NVIDIA laptops" note, that would probably be good too. And actually, it could probably go all the way back to 11th gen, since the last (I think?) non-switchable NVIDIA + Intel system was probably bonw14, which was 10th gen.

ahoneybun:
This is adding currently shipping models which does not include the `oryp11` but I will add the `addw3`.

content/graphics-switch-ubuntu.md
```diff
@@ -27,11 +27,13 @@ tableOfContents: true
...
 Many modern laptops with NVIDIA graphics cards have switchable graphics, which allows users to switch their primary GPU between the CPU's integrated graphics processor and the dedicated NVIDIA graphics card. The following System76 laptops have these switchable graphics capabilities:
 
-- Adder WS (addw1, addw2)
+- Adder WS (addw1, addw2, addw3)
+- Bonobo WS (bonw15)
 - Galago Pro (galp5 - NVIDIA models only)
-- Gazelle (gaze14, gaze15, gaze16, gaze17)
+- Gazelle (gaze14, gaze15, gaze16, gaze17, gaze18)
 - Kudu (kudu6)
 - Oryx Pro (oryp4, oryp4-b, oryp5, oryp6, oryp7, oryp8, oryp9, oryp10)
```
leviport, line 35: Looks like this line needs `oryp11` as well

leviport (approved):

thomas-zimmerman (approved):
LGTM

PR #1154, opened 2023-07-18
Update Other OS steps based on feedback

ahoneybun:


leviport (approved):

thomas-zimmerman (approved):
Good to be safe and updated before installing new software

PR #1157, opened 2023-08-01
Move article to archived

ahoneybun:
Addresses feedback on issue 1150 since the article is not getting update and the information is in the blog now.


jacobgkau:
The build is failing with this:

> Error:  Error generating route "/articles/roadmap": Article not found

Just to be clear, moving it to the "archived" folder should mean it's not being generated at all, right? (As opposed to just setting to "hidden," which would mean it's still available from the search box?)

ahoneybun:
> The build is failing with this:
> 
> > Error:  Error generating route "/articles/roadmap": Article not found
> 
> Just to be clear, moving it to the "archived" folder should mean it's not being generated at all, right? (As opposed to just setting to "hidden," which would mean it's still available from the search box?)

Correct, the article would appear in the search box even if it is disabled though will not appear on the page itself.

burden:
@ahoneybun quick heads up. Still troubleshooting, but I noticed there are links to this page in a couple spots at: `content/difference-between-pop-ubuntu.md`

content/difference-between-pop-ubuntu.md
```diff
@@ -25,7 +25,7 @@ Pop!\_OS has evolved quite a bit since its 17.10 release. While the easiest way
 
 This is a common question that comes up, and one that makes our engineers cringe. Yes, Pop!\_OS has been designed with vibrant colors, a flat theme, and a clean desktop environment, but we created it to do so much more than just look pretty. (Although it does look very pretty.)
 
-To call it a re-skinned Ubuntu brushes over all of the features and quality-of-life improvements that Pop!\_OS developers work diligently to create. For an in-depth look at the effort and manpower that goes into updating and maintaining Pop!\_OS, take a look at our [Roadmap](/articles/roadmap) documentation, you will find a general list of improvements that make Pop!\_OS stand out.
+To call it a re-skinned Ubuntu brushes over all of the features and quality-of-life improvements that Pop!\_OS developers work diligently to create.
```
jacobgkau: You could add https://support.system76.com/articles/difference-between-pop-ubuntu/ here as an alternative link to the roadmap that was here before, although that article could use an update sometime soon too.

jacobgkau (approved):
This completely removes the outdated article. We can always move it back if we decide we want to start using it again.

PR #1158, opened 2023-08-02
Create azerty-keyboard-layout tutorial

cryptolok:


ahoneybun:
Does it make sense to make this a general article about having different keyboard layouts rather than focus just on AZERTY? This is a common question for folks who are either outside the US or just like a different layout.

PR #1163, opened 2023-08-17
Add note about new 90% default start threshold

leviport:
I'm not sure if the note needs to be more obvious, but this at least gets the info into this article.

Also corrected two instances of periods inside of parentheses. They should be outside.

Side note: further down, the "Proprietary firmware" section doesn't apply to the Pangolin. I might need to file an issue about that one.

### DIFF
--- a/content/laptop-battery-thresholds.md
+++ b/content/laptop-battery-thresholds.md
@@ -16,11 +16,13 @@ tableOfContents: true
 
 Charging thresholds allow your System76 laptop to avoid charging the battery until it has dropped below a lower bound (the start threshold), and to stop charging when it reaches an upper bound (the end threshold). This is useful when your laptop is plugged into an AC power adapter for extended periods of time, as it prevents unnecessary micro-charging that would reduce battery longevity.
 
+The default charging profile for System76 laptops with Open Firmware has a start threshold of 90 and a stop threshold of 100. This increases battery longevity while keeping the battery very close to a full charge.
+
 To determine if your laptop has Open Firmware or proprietary firmware, see [this article](/articles/open-firmware-systems). (If a system has Open Firmware, then it must also have Open EC to work with charging thresholds.) See [Charging Thresholds](#configuring-charging-thresholds-open-firmware) for Open Firmware systems or [FlexiCharger](#configuring-flexicharger-proprietary-firmware) for proprietary firmware systems.
 
 ## Configuring Charging Thresholds (Open Firmware)
 
-**Note:** This feature is not currently finished. Currently, the thresholds are reset when the EC is reset (which happens when the system is shut down and the power is unplugged.) Once the feature is complete, the thresholds will be persistent and a GUI will be available to set them. To work around this limitation in the short term, you can [use systemd to set thresholds at boot](#at-boot).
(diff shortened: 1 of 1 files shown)


content/laptop-battery-thresholds.md
```diff
@@ -16,11 +16,13 @@ tableOfContents: true
 
 Charging thresholds allow your System76 laptop to avoid charging the battery until it has dropped below a lower bound (the start threshold), and to stop charging when it reaches an upper bound (the end threshold). This is useful when your laptop is plugged into an AC power adapter for extended periods of time, as it prevents unnecessary micro-charging that would reduce battery longevity.
 
-To determine if your laptop has Open Firmware or proprietary firmware, see [this article](/articles/open-firmware-systems). (If a system has Open Firmware, then it must also have Open EC to work with charging thresholds.) See [Charging Thresholds](#configuring-charging-thresholds-open-firmware) for Open Firmware systems or [FlexiCharger](#configuring-flexicharger-proprietary-firmware) for proprietary firmware systems.
+The default charging profile for System76 laptops with Open Firmware have a start threshold of 90 and a stop threshold of 100. This increases battery longevity, while keeping the battery very close to a full charge.
```
jacobgkau, line 19: "profile" is singular, so it should be "has" rather than "have."

I don't think a comma before "while" is necessary in this case, but I don't have a problem with it stylistically.

content/laptop-battery-thresholds.md
```diff
@@ -16,11 +16,13 @@ tableOfContents: true
...
+
+To determine if your laptop has Open Firmware or proprietary firmware, see [this article](/articles/open-firmware-systems). (If a system has Open Firmware, then it must also have Open EC to work with charging thresholds). See [Charging Thresholds](#configuring-charging-thresholds-open-firmware) for Open Firmware systems or [FlexiCharger](#configuring-flexicharger-proprietary-firmware) for proprietary firmware systems.
```
jacobgkau, line 21: In this case, the period actually should be inside of the parenthesis (as it was) because it's a full sentence inside of the parenthesis. It can get stylistically subjective when it comes to longer or independent clauses, but there's not really any room for interpretation on this line.

leviport: I see. Is having an entire sentence in parens correct? I'm not sure that's something I've ever seen anywhere else.

leviport: Did some research, and I guess it's correct. I'll change it.

jacobgkau: It's grammatically valid.

https://www.yourdictionary.com/articles/punctuation-inside-outside-parentheses
https://grammarist.com/punctuation/period-before-or-after-parentheses/
https://www.utsa.edu/twc/documents/Parentheses.pdf

jacobgkau (approved):

content/laptop-battery-thresholds.md
leviport: Corrected. I just removed the comma.

jacobgkau:
I assume we'll want to wait to merge this until the EC change is at least merged into EC master.

content/laptop-battery-thresholds.md
leviport: Changed it back, and learned something new today.

jackpot51 (approved):

Issue #1161, opened 2023-08-16
Mention Disks app for changing encryption key in docs


paulHempel:
https://github.com/system76/docs/blob/master/content/advanced-luks.md
This tutorial only shows the command line way. 

The 'Disks' app (which is preinstalled) provides this option as well: select disk > select encrypted partition >  settings button > 'change passphrase'.

Therefore my suggestion is to additionally mention the GUI way for non-technical users finding this docs page.

leviport:
That doc seems like a bit of an oddball. The article description in https://github.com/system76/docs/blob/master/content/password.md mentions changing LUKS passwords, but the article itself does not actually cover that. It might streamline things if we just add a LUKS section to the password.md article, then put this content into that section and get rid of the advanced-luks.md article entirely. @ahoneybun what do you think?

ahoneybun:
I believe there was some possible data lost with changing it which makes me nervous to even keep it without a disclaimer.

leviport:
I'm pretty sure that bug was solved long ago.

PR #1166, opened 2023-09-07
Work on adding the advanced luks article to the password article

ahoneybun:
This combines them fixing issue #1161


thomas-zimmerman (commented):
...and back to Aaron to make sure he likes my edits.

content/password.md
```diff
@@ -114,7 +114,7 @@ The output may be different based on the drive setup and partition table. You ca
 sudo cryptsetup luksDump /dev/sda3
 ```
 
-Replacing '/dev/sda3' with the location of the root partition on your system. With the output of this command we can see the seven extra slots that we have for passwords to decrypt the drive. Slots 1-7 are the open ones and Slot 0 is the current one the system has set.
+Replacing '/dev/sda3' with the location of the root partition on your system. With the output of this command we can see the seven extra slots that we have for passwords to decrypt the drive. On a normal install Slots 1-7 are the open, with Slot 0 as the initial pass-phrase.
```
ahoneybun, line 117: This is a bit odd:

On a normal install Slots 1-7 are the open, with Slot 0 as the initial pass-phrase.

I feel like it would be like this:

On a normal install Slots 1-7 are not used and Slot 0 is used as the initial pass-phrase.

ahoneybun (commented):
Why are you using "pass-phrase" instead of just "passphrase"?

thomas-zimmerman (approved):
Let's get this live!

content/password.md
```diff
@@ -30,11 +32,11 @@ First, bring up the <u>systemd-boot</u> menu by holding down <kbd>SPACE</kbd> or
...
 
-## Ubuntu
+### Ubuntu
 
 If you are using Ubuntu we can boot from a live disk and [here](/articles/live-disk) are instructions for creating the live disk. Once the live disk has been created refer to this [article](/articles/boot-menu) for booting from the live disk.
 
-## Mounting the Installed OS
+#### Mounting the Installed OS
 
 Press <kbd><font-awesome-icon :icon="['fab', 'pop-os']"></font-awesome-icon></kbd>/<kbd><font-awesome-icon :icon="['fab', 'ubuntu']"></font-awesome-icon></kbd>+<kbd>T</kbd> to open a terminal, and type in these commands:
```
XV-02, line 41: Most systems don't use either the Ubutu or Pop themed super keys, including every currently availiable System76 product baring the Launch Heavy. Beyond this, ``super + T `` doesn't yield a terminal under Ubuntu, where the short cut is ``ctrl + alt + T``.

XV-02 (commented):
There are several places where we assume that a pop install will be on certain devices/ partitions (nvme0n1p3 or sda3) which could cause issues. An image inline, highlighting the listed partition may help a user. More importantly, if we assume that our user is not technically inclined, parts of the article which haven't been touched cite commands using "sda3" when A) "nvme0n1p3" is probably more common now given how many major manufacturers have moved to default nvme drives, let alone our own nvme-centric offerings, and B) might cause users to hit a wall where they just type commands in character-for-character without understanding them. Admittedly, a support article will never replace someone actively engaging with the instructions... but it might still be worth considering changes to some of that verbiage as we're already looking at updating this article.

(I've yet to review the LUKS portion, but will be doing that next)

content/password.md
```diff
@@ -84,4 +86,52 @@ exit
...
+## Adding LUKS passphrases
+
+Full Disk Encryption supports up to 8 total pass-phrases to unlock the encryption. This is useful if you would like to allow more than a single user to unlock the encryption and not share a pass-phrase. The default pass-phrase is in slot 0, with space for additional pass-phrases in slots 1 through 7 (for a total of 8).
+
+### List Partitions
+
+First let's list the partitions of all of the installed drives:
+
+```
+lsblk -f
```
XV-02, line 108: While it is more to type, I would be inclined to replace `lsblk -f` with `lsblk -o name,mountpoints` as it strips out a large quantity of unnecessary information, and will not cause incredibly hard-to-parse line-wrapping to occur even on smaller terminal windows (like the default terminal window size when spawned as a floating window)

content/password.md
```diff
@@ -84,4 +86,52 @@ exit
...
+```
+
+The output may be different based on the drive setup and partition table. You can see the root partition at mountpoint '/'. From there we can use that output to run this command:
+
+```
+sudo cryptsetup luksDump /dev/sda3
```
XV-02, line 114: I'm uncertain how, to the uninitiated, this command demonstrates that there are 7 extra slots. It might be worth highlighting how, under the "Keyslots:" section, a single entry "0" should be seen, or to otherwise (and hopefully more elegantly that my suggestion) draw attention to that specific aspect.

content/password.md
XV-02, line 108: Change `lsblk -f` to `lsblk -o name,mountpoint`

The amount of text garbage - for the average end user, and in the context of our needs here - this change cuts out is well worth it. It makes the output much easier to display, to read, and to understand. I actually do consider this a pretty important change.

XV-02 (changes requested):

XV-02 (approved):

thomas-zimmerman (approved):
Let's get this change out to the world

PR #1169, opened 2023-09-11
Add link to pang13

ahoneybun:


jacobgkau (changes requested):
This should be called Pangolin, not Pangolin Pro. Looking at tech docs and Joshua reports, I believe that should be the case going back to the pang10.

ahoneybun:
> This should be called Pangolin, not Pangolin Pro. Looking at tech docs and Joshua reports, I believe that should be the case going back to the pang10.

Should be fixed now.

jacobgkau (approved):

thomas-zimmerman (approved):
LGTM - let's get this live

PR #1171, opened 2023-09-12
New Fan Noise article

ahoneybun:
This adds a new article about fan noise to help explain the behavior that customers are seeing and ways to help if it is too loud for personal preference.


content/fan-noise.md
```diff
@@ -0,0 +1,36 @@
+---
+title: System Fan Noise and Expectactions
+description: >
+  Ways to manage fan noise what what to expect
```
leviport, line 4: This sentence is weird

content/fan-noise.md
```diff
@@ -0,0 +1,36 @@
...
+twitterImage: /_social/article
+
+hidden: false
+section: software
+tableOfContents: true
+---
+
+## Noise Levels
+
+We set custom fan curves for some laptops to maximize their performance so that you get as much power as possible but still keep the system in safe temperature ranges.
```
leviport, line 19: This sentence is a run-on

content/fan-noise.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Monitoring Fan Speeds
+
+Fan Speeds can be reviewed in the <u>psensor</u> which can be installed from either the <u>Pop!_Shop</u> or the <u>Ubuntu Software</u> applications. Here is what it should look like (the number of cores/threads will be different depending on the system as well as having a GPU fan or not).
```
leviport, line 23: Are the fan speeds being reviewed or just viewed?

content/fan-noise.md
```diff
@@ -0,0 +1,36 @@
...
+
+![psensor on an oryp7](/images/fan-noise/psensor-oryp7.png)
+
+### Fan Modes
+
+Most laptops other then the Pangolin (pang12 and pang13) have fan modes which are max or automatic. Changing between these modes can be done with the Fn+1 key combo (if you model supports this it will be in their [tech-docs](https://tech-docs.system76.com)).
```
leviport, line 29: You want "other than", not "other then".

leviport (changes requested):

thomas-zimmerman:
I worked on the language for the article. I think we would want to have a bit more than work on battery or turn on battery life mode. 

Expanding to show how to customize the firmware would likely be the next step? Highlighting the advantages of open firmware is good but I have slight worries about needing to support bad firmware builds. Altering the fan curve should not be difficult but the chance of soft-bricking a device is very scary.

leviport:
Firmware is pretty intensive, so I'm not sure we want to mention it in support docs. At least not without massive warnings and disclaimers about what happens if you botch it.

Mentioning something about user-controlled fan curves being planned some time in the future would probably be sufficient.

content/fan-noise.md
```diff
@@ -0,0 +1,36 @@
...
+twitterImage: /_social/article
+
+hidden: false
+section: software
+tableOfContents: true
+---
+
+## Noise Levels
+
+System76 sets custom fan curves for laptops to maximize their performance while minimazing fan noise. Current firmware keeps the fan off until 65c and will ramp the fan speeds to max speed by 90c.
```
leviport, line 19: s/minimazing/minimizing/g

s/65c/65°C/g

s/90c/90°C/g

content/fan-noise.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Monitoring Fan Speeds
+
+Fan Speeds can be monitored with <u>psensor</u>, which can be installed from either the <u>Pop!_Shop</u> or the <u>Ubuntu Software</u> applications. Below is an example from a Oryx Pro (the number of cores/threads/GPU presense will be different depending on the system).
```
leviport, line 23: s/presense/present/g

content/fan-noise.md
```diff
@@ -0,0 +1,36 @@
...
+
+![psensor on an oryp7](/images/fan-noise/psensor-oryp7.png)
+
+### Fan Modes
+
+Our Open Firmware laptops (so other than the Pangolin) have fan controls to swap between max or automatic thermal control. Use Fn+1 to toggle between the modes. (Your model will show support on the key section of their [tech-docs](https://tech-docs.system76.com)).
```
leviport, line 29: >(so other than the Pangolin)

This doesn't work for me. Older laptops also don't have open firmware. If you want to be specific, it'd probably be best to link to https://github.com/system76/firmware-open/blob/master/FEATURES.md or something. Otherwise, just delete this part. 

>will show support on the key section

Be more specific than "the key section". Something like `the "Keyboard Shortcuts" section in the External Overview` would be much more specific.

content/fan-noise.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Easy settings to effect fan noise
+
+Fan speed is based on termal output of the system. There are some settings that will indirectly limit how much heat is created during use:
```
leviport, line 33: s/termal/thermal/g

content/fan-noise.md
```diff
@@ -0,0 +1,36 @@
...
+
+1. Run the laptop while on battery: This will limit power budget for the CPU and GPU and will lower thermal output.
```
leviport, line 35: >power budget

This is the first time I've heard this terminology. Where did it come from?

content/fan-noise.md
```diff
@@ -0,0 +1,36 @@
...
+2. Set the laptop to Battery Life mode in either Power Profiles in the system menu or using `system76-power`: This limits the turbo mode on the CPU and dim the display to prolong battery life. A side effect is to also limit how much heat is created.
```
leviport, line 36: >system menu or using `system76-power`: This

Sentence should end with a period, rather than a colon.

>limits the turbo mode on the CPU

"Limiting the CPU's boost frequency" is a better way of describing this.

>and dim the display

Should be "dims the display", but is dimming the display relevant to the subject matter? Might make sense to drop that part.

leviport (changes requested):

content/fan-noise.md
ahoneybun: I believe this is from @thomas-zimmerman ?

content/fan-noise.md
ahoneybun: I believe for the display part that would increase the battery life.

content/fan-noise.md
leviport: I think we should change it to "power consumption" instead. It describes what's being limited more directly and clearly.

leviport (approved):
The open firmware chart changes seem out-of-scope for a "new fan noise article" but they still look like good changes to me, so I approve.

thomas-zimmerman (approved):
decent base article. Need to start collecting other ways to improve fan noise on Open Firmware systems.

PR #1172, opened 2023-09-13
Add step to ensure power is fully drained from components

winfred:
My thelio can get in this state even on a normal/safe shutdown sometimes. Has happened infrequently enough that I forget this crucial step each time haha


jacobgkau:
This was discussed somewhat thoroughly in https://github.com/system76/docs/pull/1053.

This PR is better since it doesn't unnecessary state to repeat the procedure multiple times. Although it still lists the power switch being on before it suggests unplugging the system, that order of troubleshooting makes sense based on how commonly each step might be needed.

I do want to clarify, when you say `but you see lights on`, is that the front power button, or lights inside of the machine? If it's the former, then we might want to move this underneath step 5. (Or we could make this more general by removing that note.)

PR #1174, opened 2023-09-15
fix builds for node 18+

burden:
tested:
- v18.17.1
- v20.6.1


ahoneybun (approved):
Tested the following so far:

- Images load
- Internal links work
- Code blocks are good

macifell (approved):
:+1:

PR #1176, opened 2023-09-20
Update ubuntu articles

ahoneybun:
This handles some minor updates to a few Ubuntu focused articles with updated information.


content/graphics-switch-ubuntu.md
```diff
@@ -102,6 +102,8 @@ Now enable the System76 Power extension with the toggle to the right.
 Click the system menu in the top right corner of your screen to access graphics
 switching.
 
+>**Note**: Ubuntu versions higher then 22.04 will not have this menu as the GNOME Shell extension has not been updated for newer GNOME versions.
```
leviport, line 105: "higher then" should be "higher than"

leviport (changes requested):

content/graphics-switch-ubuntu.md
jacobgkau: I'd recommend referencing (and linking to, if possible) the command-line section within this note to present the alternative to what we're saying isn't an option.

I'd also replace "as" with "because".

leviport (approved):

content/graphics-switch-ubuntu.md
```diff
@@ -102,6 +102,8 @@ Now enable the System76 Power extension with the toggle to the right.
 Click the system menu in the top right corner of your screen to access graphics
 switching.
 
+>**Note**: Ubuntu versions higher than 22.04 will not have this menu because the GNOME Shell extension has not been updated for newer GNOME versions.
```
thomas-zimmerman, line 105: I really dislike how this is worded. Better would be:
**Note**: Gnome Shell versions newer than 42 are not supported by the System76 Power extension

ahoneybun: I feel like the issue is that most folks don't know the version of GNOME that ships with their version of Ubuntu. Maybe this?:

GNOME Shell versions newer than 42 (Ubuntu 22.04 and above) are not supported by the System76 Power extension.

XV-02: Ubuntu 22.04 ships with a Gnome 42 implementation, so:

**Note:** Gnome Shell versions newer than 42 (Ubuntu 22.10 or later) are not supported by the System76 Power extension.

ahoneybun: New commit changes this to match that.

leviport (approved):

PR #1178, opened 2023-09-26
Update dock article

ahoneybun:
This updates the content and screenshots including the new steps for Ubuntu 22.04.


thomas-zimmerman (dismissed):
npm live view looks good.

thomas-zimmerman (approved):
Nice update to the article

content/customize-dock.md
```diff
@@ -0,0 +1,114 @@
...
+
+At this point, Firefox can be closed. The dock will display on the bottom of the screen. You will want to disable COSMIC Dock in the <u>Extension</u> application as well.
+
+![Pop!_OS Extension application](/images/customize-dock/pop-gnome-extensions.png)
+
+To customize the dock's settings, click on the settings button for Dash to Dock in the <u>Extension</u> application"
+
+![Dash to Dock Settings](/images/customize-dock/pop-dash-to-dock-settings.png)
+
+In the Dash to Dock Settings window, you can choose which screen edge to display the launcher on, such as the bottom of the screen. You can also choose to extend the dock into "Panel mode," which looks similar to Ubuntu:
```
leviport, line 62: >edge to display the launcher on

Probably want to change "launcher" to "dock" here. Launcher is still something else.

content/customize-dock.md
```diff
@@ -0,0 +1,114 @@
...
+
+Ubuntu includes a lightweight version of Dash to Dock, which is why the dock already displays on the left side of the screen by default.
+
+![Stock Ubuntu Dock](/images/customize-dock/ubuntu-dock.png)
+
+If you need more options or want to make more changes to the dock you can install Dash to Dock using the steps below.
+
+## Dash to Dock on Ubuntu
+
+When using Dash to Dock on Ubuntu it can effect the default Dock as it was forked from that project. When you enable the full version of Dash to Dock, it will override the built-in version.
```
leviport, line 74: s/effect/affect/g

content/customize-dock.md
```diff
@@ -0,0 +1,114 @@
...
+
+To install Dash to Dock on Ubuntu, you will first need to install the "native host connector," which allows Firefox to install and remove GNOME Shell extensions and the GNOME Shell <u>Extension</u> application. This component is installed by default in Pop!\_OS, but it is not installed by default in Ubuntu. Open a terminal and run this command to install the native host connector:
```
leviport, line 76: >the GNOME Shell <u>Extension</u> application

s/Extension/Extensions/g

content/customize-dock.md
```diff
@@ -0,0 +1,114 @@
...
+
+After refreshing the web page, you should see an On/Off switch at the top right of the page. Click the switch to toggle it to the "On" position. On the resulting pop-up, click Install to download and install the extension.
+
+![Installing the GNOME Shell extension](/images/customize-dock/ubuntu-install-dash-to-dock3.png)
+
+At this point, Firefox can be closed. The dock will now display on the bottom of the screen, but not in Panel mode. You will want to disable Ubuntu Dock in the <u>Extensions</u> application:
+
+![Ubuntu Extension application 1](/images/customize-dock/ubuntu-gnome-extensions1.png)
+
+Then enable Manually Installed extensions and then enable Dash to Dock:
```
leviport, line 104: >Then enable Manually Installed extensions 

Should this be "scroll to Manually Installed"? It doesn't look like there's anything to enable in the "Manually Installed" section in that screenshot.

leviport (changes requested):

content/customize-dock.md
ahoneybun: This looks fine to me in the live version.

Screenshot from 2023-09-28 07-38-41

content/customize-dock.md
```diff
@@ -0,0 +1,114 @@
...
+
+Pop!\_OS and Ubuntu both contain an application dock on the left side of the screen. You can adjust this preference, move the dock to the bottom of the screen, and more in <u>Settings</u>.
+
+## Default Dock on Pop!\_OS
+
+Pop!\_OS includes a lightweight version of Dash to Dock, which is why the dock already displays on the bottom of the screen by default.
+
+![Stock Pop Dock](/images/customize-dock/pop-dock.png)
+
+If you need more options or want to make more changes to the dock you can install Dash to Dock using the steps below.
```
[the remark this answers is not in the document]
ahoneybun: Wouldn't this also be the same for Ubuntu? Do we just want to get rid of Dash to Dock completely?

content/customize-dock.md
leviport: I can't figure out what the sentence is telling me to do. If I can't figure it out, someone else is going to struggle to understand it as well. It tells me to enable two different things, and I'm not sure what the first thing is that it wants me to enable.

content/customize-dock.md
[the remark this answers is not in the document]
leviport: Dash to Dock is probably less likely to have compatibility issues on Ubuntu, but I guess that would be fine. Your call, I guess.

content/customize-dock.md
ahoneybun: Ah I think I see the issue, I think the Extensions at the top was disabled which caused Dash to Dock to be greyed out.

content/customize-dock.md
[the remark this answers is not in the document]
ahoneybun: I don't see any options for COSMIC Dock in dconf only dash-to-dock stuff so I don't think we expose those options. So perhaps Dash to Dock is better when you need more options.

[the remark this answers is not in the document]
leviport: They're under dash-to-dock in dconf-editor. They're all there.

ahoneybun:
I made changes to remove Dash to Dock installation and use dconf Editor instead per @leviport 's suggestion.

content/customize-dock.md
```diff
@@ -0,0 +1,48 @@
...
+hidden: false
+section: software
+tableOfContents: true
+
+redirect_from:
+  - /movelauncher
+  - /articles/move-launcher
+---
+
+Pop!\_OS and Ubuntu both contain an application dock on the bottom of the screen. You can adjust this preference, move the dock to the bottom of the screen, and more in <u>Settings</u>.
```
leviport, line 26: >Pop!\_OS and Ubuntu both contain an application dock on the bottom of the screen.

It's on the left side for Ubuntu. I think they should either have their locations mentioned separately so that they are both correct, or the locations should be omitted entirely.

>move the dock to the bottom of the screen,

If the previous sentence says it's already on the bottom edge of the screen, it doesn't make sense to say that it can be moved to the bottom of the screen (it's already there). Something like "change the location of the dock" seems like it would fit better.

content/customize-dock.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Default Dock on Pop!\_OS
+
+Pop!\_OS includes a lightweight version of Dash to Dock, which is why the dock already displays on the bottom of the screen by default.
+
+![Stock Pop Dock](/images/customize-dock/pop-dock.png)
+
+## Default Dock on Ubuntu
+
+Ubuntu includes a lightweight version of Dash to Dock, which is why the dock already displays on the left side of the screen by default.
```
leviport, line 36: I think Dash to Dock sets the location to the bottom edge by default, so saying that's why it's on the left edge seems confusing.

leviport (changes requested):

content/customize-dock.md
```diff
@@ -23,17 +24,17 @@ redirect_from:
   - /articles/move-launcher
 ---
 
-Pop!\_OS and Ubuntu both contain an application dock on the bottom of the screen. You can adjust this preference, move the dock to the bottom of the screen, and more in <u>Settings</u>.
+Pop!\_OS and Ubuntu both contain an application dock on the screen. You can adjust this preference, such as move the dock to a different side the screen, and more in <u>Settings</u>.
```
leviport, line 27: >You can adjust this preference, such as move the dock to a different side the screen, and more in <u>Settings</u>.

"such as" makes this sentence not work, and it's unclear what "this preference" refers to, since the position was removed from the previous sentence.

If I were to rewrite it, I'd go with something like this:

>The location of the dock, as well as other dock settings, can be adjusted in Settings.

content/customize-dock.md
```diff
@@ -23,17 +24,17 @@ redirect_from:
...
 
 ## Default Dock on Pop!\_OS
 
-Pop!\_OS includes a lightweight version of Dash to Dock, which is why the dock already displays on the bottom of the screen by default.
+Pop!\_OS includes a lightweight version of Dash to Dock like in the screenshot below:
```
leviport, line 31: "the screenshot below" wasn't mentioned before. Why was that added? 

Also I don't think either COSMIC Dock or Ubuntu's dock are any lighter than Dash to Dock, so I'm not so sure "lightweight" has any basis in this note or the one below. But I've been overlooking that on my past reviews, so if you want to leave it, I won't complain too much. I'm just uncertain about it.

leviport (changes requested):

leviport (approved):

content/customize-dock.md
```diff
@@ -0,0 +1,49 @@
...
+redirect_from:
+  - /movelauncher
+  - /articles/move-launcher
+---
+
+Pop!\_OS and Ubuntu both contain an application dock on the screen. The location of the dock, as well as other dock settings, can be adjusted in <u>Settings</u>.
+
+## Default Dock on Pop!\_OS
+
+Pop!\_OS includes a lightweight version of Dash to Dock.
```
leviport, line 31: I already approved this PR, but I think it might be to change this to: 

>Pop!_OS's dock is positioned along the bottom edge of the screen by default.

Just because "lightweight" still doesn't make sense to me and the sentence feels like a placeholder. Up to you whether or not you want to implement it. Obvs. you'd want to make the Ubuntu one below match this one.

ahoneybun: Should be updated now so we can get another approval on this PR.

thomas-zimmerman (approved):
LGTM

PR #1180, opened 2023-09-27
Log in from live disk

thomas-zimmerman:
add additional context for the commands that un-encrypt the OS drive


content/login-from-live-disk.md
```diff
@@ -19,7 +19,7 @@ section: software-troubleshooting
 tableOfContents: true
 ---
 
-## Login from Live Disk (Chroot)
+## Log in from Live Disk (Chroot)
```
XV-02, line 22: If we're changing "Login" to "Log in" here, we should also change it in the article title for consistency's sake.

ahoneybun (approved):
Good changes.

leviport (approved):

XV-02 (approved):

PR #1181, opened 2023-09-29
Update screenshots and content for 22.04

ahoneybun:


XV-02 (changes requested):
`sudo apt install ./Lightworks*.deb` should be `sudo apt install ./lightworks*.deb` with a lower-case "L". 
I don't know if LWKS are good about being consistent with their package name, but it currently is lower-cased :man_shrugging:

XV-02 (approved):

fs1995 (approved):

PR #1182, opened 2023-10-06
Change table to command to check hardware

ahoneybun:


content/login-loop-pop.md
```diff
@@ -120,18 +120,11 @@ After reinstalling those packages and rebooting, try logging in again.
...
-| Serval WS     | Galago Pro 5   | Meerkat     |
-| Bonobo WS     | Gazelle        | Galago Pro (4 or older)  |
-| Leopard WS    | Thelio         | Darter Pro  |
-| Silverback WS | Thelio Mira    | Pangolin    |
-| Thelio Mega   | Thelio Major   |             |
-|               | Thelio Massive |             |
+If your system is equipped with NVIDIA graphics, a recent update might be causing the login issues. (Usually, NVIDIA driver issues will prevent the login screen from loading at all; however, it's still worth reinstalling if you are unable to log in after the above steps.) Run these commands to determine if your system has NVIDIA graphics:
+
+```bash
+lspci -v | grep VGA
```
XV-02, line 126: At current, we ship with `ubuntu-drivers-common` which provides the `nvidia-detector` command. This will scan for nvidia devices, and either return the correct driver version for hardware, or otherwise return "none". 

This might be a better command as it will have less text output than `lspci -v |grep VGA` and doesn't require a user - who is likely in a distressed situation - to parse between Intel, AMD, or Nvidia devices in what may be unfamiliar output.

thomas-zimmerman (approved):
LGTM

XV-02 (approved):

PR #1185, opened 2023-10-06
Add meer7 and darp9

ahoneybun:
This also adds the darp9 rather then having it as a different PR since it moves the table.


jacobgkau (approved):
Displaying as expected & the links work.

image

fs1995 (approved):

PR #1186, opened 2023-10-06
Update launch_2-firmware-update.md

garrettjwilke:
added clarification on which launch keyboards are affected with this firmware update


leviport:
launch_1 and launch_lite_1 can still be flashed with this method, so I don't think the distinction is needed

PR #1189, opened 2023-10-16
Remove duplicate instruction

mapachurro:
The instruction regarding taking a picture or screenshot of the instructions showed up twice. 

I left the later one, as I feel it provided better emphasis to an important point.


leviport:
Thanks for catching that! I'm kinda thinking the second one should be removed in favor of the first one. Having the instructions is important, but the other things mentioned are probably equally important, and it looks more organized if all the notes are together.

@ahoneybun what do you think?

ahoneybun:
@leviport like that?

leviport:
Close, but that broke the linter and left an extra asterisk.

ahoneybun (approved):

PR #1190, opened 2023-10-18
Update desktop-environment.md

memachado:
This addition offers users a more comprehensive solution for removing KDE and its associated packages. This enhancement aims to provide clearer and more user-friendly instructions for managing desktop environments on Pop!_OS.

### DIFF
--- a/content/desktop-environment.md
+++ b/content/desktop-environment.md
@@ -23,7 +23,7 @@ Pop!\_OS and Ubuntu both include the GNOME desktop environment by default. A des
 
 You can install an alternative desktop environment using the instructions below.
 
-**NOTE:** be careful when installing other desktop environments, as they may affect the default GNOME desktop (both Ubuntu and Pop).
+> **Warning:** Installing other desktop environments may affect the visual style and functionality of the default GNOME desktop (on both Ubuntu and Pop!_OS).
 
 If you run into trouble while using an alternative desktop environment, you may wish to revert to the default environment. To ensure the default GNOME desktop environment is installed in Pop!\_OS, install the `pop-desktop` package:
 
@@ -229,16 +229,16 @@ gsettings set org.cinnamon.desktop.lockdown disable-lock-screen false
 
 ### Removing Desktop Environments
 
-If you no longer want to use a desktop environment, it can be removed by using:
+If you no longer want to use a desktop environment, it can be removed by using the `sudo apt autoremove --purge` command with the name of the package you originally installed. For example, if you installed LXDE with `sudo apt install lxde`, you can remove it using the following command:
 
 ```bash
-sudo apt autoremove --purge ...
+sudo apt autoremove --purge lxde
 ```
 
-For example, to remove KDE:
+Some desktop environments (such as KDE) may leave behind additional packages even after uninstalling the original package with the `autoremove` command. To more thoroughly remove packages related to a desktop environment, use a regular expression to remove all packages beginning with the name of the environment followed by a hyphen (for example, `kde-`):
+
+> **Warning:** Because regular expressions can match packages you don't expect, inspect the list of packages being removed before confirming the operation, and be prepared to reinstall any packages you wish to keep afterwards.
 
 ```bash
-sudo apt purge kde-*
-
-sudo apt autoremove --purge kde-standard
+sudo apt autoremove --purge 'name?(^kde-)'
 ```
--- a/content/desktop-environment.md
+++ b/content/desktop-environment.md
@@ -238,5 +238,7 @@ sudo apt autoremove --purge ...
 For example, to remove KDE:
 
 ```bash
+sudo apt purge kde-*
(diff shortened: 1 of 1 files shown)


jacobgkau (changes requested):
The article shows how to install KDE by running `sudo apt install kde-standard`, which will install `kde-standard`'s dependencies. Running an `autoremove --purge kde-standard` will remove `kde-standard` and everything installed as a dependency of it. There's no reason why `kde-*` would be needed unless the user installed KDE via a different package/method.

The `kde-*` wildcard may uninstall more packages than intended; for example, if I installed the `kde-spectacle` program to use in GNOME, then I later installed `kde-standard`, running `sudo apt purge kde-*` will remove `kde-spectacle` even though it wasn't installed as part of the desktop environment.

Also, `purge` is actually less comprehensive than an `autoremove --purge` because `purge` will not remove dependencies by itself; it would require another `autoremove --purge` after it. (In this case, there is one immediately after it, but specifying `kde-standard` in that one after `kde-*` was already purged is redundant.) It's probably best to use `autoremove --purge` in all instances here.

I think there's a place for the wildcard/pattern-matching concept in this article, but it should be separated into a separate section for if the reader's forgotten which package(s) they manually installed.

Additionally, * wildcards aren't reliable; if you `touch kde-testing` and then try to run the `sudo apt purge kde-*` command in the same directory, you'll get an error that no `kde-testing` package is found, because that wildcard is matched to the filename rather than package names. Instead, `sudo apt purge '?name(^kde-)'` would do the same thing (remove packages starting with `kde-`), without running into issues in case of matching filenames.

memachado:
@jacobgkau  Thank you for the feedback and explanation. I did not consider those points prior to the PR. I'll run some more tests, if nothing goods come out, I drop the PR. Bye.

memachado:
@jackpot51 

To test the current example and documentation:

- Fresh installation
- Installed the KDE Standard as described in the docs
- Rebooted, logged in, and tested KDE
- Rebooted again, logged into GNOME
- "Removed" as described in the docs
- Rebooted and was able to log in again to the KDE environment because KDE was not uninstalled, hence my attempt to further enhance the docs.

The method in this PR was found on Stack Overflow, but I traced it to the old version of these docs.

Although the old/proposed method can remove unintended packages, the current method does not remove the necessary packages.

jacobgkau:
Thank you for the steps and elaborating on the issue. I can recreate that with `kde-standard`. I'm not quite sure what's going on yet, but I'll try to look into it a little further.

(Playing around with `-o APT::Autoremove::SuggestsImportant=0` and `-o APT::Autoremove::RecommendsImportant=0`, it looks like it might be other, preinstalled packages that suggest or recommend some of the KDE packages or their dependencies/recommendations. The former option removes too little and the latter option removes too much, though. `plasma-workspace` and `plasma-desktop` may be creating the issue between themselves, from what I'm seeing trying to use `aptitude why` on them.)

If we do stick with the wildcard option, we'll still want to use regex rather than the asterisk.

memachado:
> If we do stick with the wildcard option, we'll still want to use regex rather than the asterisk.

Still, the regex would need to be `sudo apt purge '?name(-kde-)'` because packages like `qml-module-org-kde-*`. And even after that some packages will remain.

Maybe change the example to another desktop environment, and include a warning about kde remaining packages after install? This way people can choose how to proceed.

jacobgkau:
> Maybe change the example to another desktop environment, and include a warning about kde remaining packages after install?

I don't see how removing the KDE example will help if that's one that we know currently has problems.

> Still, the regex would need to be `sudo apt purge '?name(-kde-)'` because packages like `qml-module-org-kde-*`. And even after that some packages will remain.

Actually, using the `^kde-` regex with `autoremove`, I do see `qml-module-org-kde-...` get removed even though it wouldn't be matched by that regex:

image

With `autoremove`, we don't actually need to match every package we want to be uninstalled, we just need to match the ones that apt doesn't want to uninstall unless we tell it to.

jacobgkau:
I've tried to explain the example in a simple but effective way. I do think this is a valuable addition, so thanks for bringing it up.

jacobgkau (approved):

memachado:
> With autoremove, we don't actually need to match every package we want to be uninstalled, we just need to match the ones that apt doesn't want to uninstall unless we tell it to.

My bad. I used `sudo apt purge '?name(^kde-)'`  instead of `sudo apt autoremove --purce '?name(^kde-)'` . 

> I've tried to explain the example in a simple but effective way. I do think this is a valuable addition, so thanks for bringing it up. (Let me know if you see any issues with what I changed it to.)

Looks good to me. 

Thank you.

ahoneybun (approved):

ahoneybun:
@memachado is this ready to go with your latest changes?

memachado:
@ahoneybun Yes.

PR #1198, opened 2023-12-06
Redo bootloader

ahoneybun:
This PR reorders sections but also updates sections with updated information and clearer steps.


content/bootloader.md
```diff
@@ -90,11 +102,17 @@ Installed in UEFI mode
...
 ### EFI Boot - Pop!_OS (systemd-boot)
 
-If the echo command at the beginning of this page says that the OS is installed in EFI mode **and** you are using Pop!_OS, follow this section. Please note that if you have an encrypted disk, you will need to first unlock it as described below.
+If the echo command at the beginning of this page says that the OS is installed in EFI mode **and** you are using Pop!_OS, follow this section. Please note that if you have an encrypted disk, you will need to first unlock it as described below. If you see the error below then your drive is encrypted with LUKS:
+
+```
+mount: /mnt: unknown filesystem type 'crypto_LUKS'.
+```
+
+Follow these [steps](/articles/bootloader#encrypted-disk) to decrypt the drive first.
```
thomas-zimmerman, line 115: Instead of repeating bounce to the set of commands for encryption, I think we should do something like:
```
sudo cryptsetup luksDump /dev/nvme0n1p3
```
If we get output that states it's not encrypted that you would skip the steps for the manual unlocking of the drive:
```
Device /dev/nvme0n1p3 is not a valid LUKS device.
```

delrioj (approved):

brandonbaez7 (approved):

PR #1200, opened 2023-12-13
Update images and comment out CLI option 

ahoneybun:
This is not supported in v4.


leviport (changes requested):
Looks good, but fix that lint. Just need to remove one of those newlines.

leviport (approved):

thomas-zimmerman (approved):
Nice change

PR #1202, opened 2023-12-19
Archive obsolete article about battery percentage setting

ahoneybun:
This moves the article to be archived since it is no longer valid information.

### DIFF
--- a/archived/show-battery-percentage.md
+++ b/archived/show-battery-percentage.md
@@ -10,7 +10,7 @@ keywords:
 facebookImage: /_social/article
 twitterImage: /_social/article
 
-hidden: false
+hidden: true
 section: community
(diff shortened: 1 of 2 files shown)


jacobgkau (approved):
I would agree that this article isn't really needed anymore if it's just detailing one setting in the default Settings app.

This is working when I test it locally (the article doesn't show up in the list or search), so I'm approving. However, you may also want to change `hidden` to `true` since that seems to be the case for most of the other articles in the `archived` directory.

jacobgkau (dismissed):
More descriptive commit messages would be nice to work on in the future.

jacobgkau:
Was that commit on the codecs article supposed to go into the same PR as the battery percentage article archival?

PR #1203, opened 2023-12-26
Update article about Differences between Pop!_OS and Ubuntu

samcodee:
- Added an updated version of download screen image
- Added a line about the Raspberry pi 4 ISO
- Fixed a minor typo in "backported", previously "backpored"
- Deleted old download screen image

this is literally my first PR, really sorry if ive done anything wrong


jacobgkau (approved):
Before:
image

After:
image

That image could probably use a frame to help make it clear the download buttons aren't actually on the page. The update is fine, though.

The typo fix lower down in the article is also valid.

We can probably rework this article from scratch sometime in the near future, since this question doesn't come up as often anymore and there are more obvious and more numerous answers than what's covered right now.

ahoneybun (approved):
Great first PR and thank you for the work!

PR #1204, opened 2023-12-26
Use recursive binding before chroot

PeterGrand:
By using -R instead of -B, we ensure that the nested mount of /sys/firmware/efi/efivars/ becomes mounted in our target filesystem, which allows us to do various boot repairs with an EFI system.


jacobgkau (approved):
Confirmed `efivars` were not available in the chroot before, and are after this change. Also took a look at both flags in the `mount` manual page. This seems like a good change.

ahoneybun (approved):

PR #1208, opened 2024-01-02
Update switch-from-macos-to-ubuntu.md

callpri:
Codelobster editor is added


jacobgkau (commented):
Codelobster doesn't seem to be open-source (seems to be shareware with a timed email registration requirement and paid options), and also doesn't seem terribly popular. Trying to search the web for information about it, some of the only discussion I found was the [PR author's Reddit account](https://old.reddit.com/user/callpri2) recommending it in many threads. @callpri For transparency's sake, do you have a connection with Codelobster, or are you just a user evangelizing it?

I'm not against having it visible in general, but I'm not sure if this article needs it mentioned. On the other hand, Sublime Text is also closed-source and paid (and not super actively developed these days). At the least, maybe we should have the new addition added to the end of the list (after Sublime instead of before)?

ahoneybun:
@jacobgkau it looks like Zed might be a better replacement for Sublime and it has it's code on Github.

PR #1209, opened 2024-01-05
Fix typo in difference-between-pop-ubuntu.md

bbarker:
Fix typo in keyboard shortcut for switching workspaces


leviport (approved):
Thanks for the correction! I think that was a holdover from the days before pop-shell, before the shortcut required Ctrl.

ahoneybun (approved):

PR #1212, opened 2024-01-10
Update guides.md to include Lemur Pro 12 (lemp12) tech doc

seansawyer:
Noticed that my model wasn't listed in `guides.md`, but it does have a tech docs page.


leviport (approved):
Thanks for the correction! Looks like we missed that one

ahoneybun (approved):

PR #1213, opened 2024-02-05
Add pang13 and pang14 to missing spots in articles

ahoneybun:
We may want to redo this a bit with the growing list of Pangolin models that these sections of articles cover.


leviport (approved):

thomas-zimmerman (approved):
Checked changes live. Not seeing any breakage. The large table on the windows article will need work -- it's not great looking atm.

PR #1221, opened 2024-03-21
Update hardware-failure.md

jklgrasso:
Changed the gpu-burn example to use tc instead of d to create more of a workload when testing.

### DIFF
--- a/content/hardware-failure.md
+++ b/content/hardware-failure.md
@@ -153,10 +153,16 @@ Now we'll compile it:
 make
 ```
 
-And now we can run it like so (this example will run it for 60 minutes/1 hour):
+For NVIDIA **RTX** GPUs we can use Tensor cores to run it like so (this example will run it for 60 minutes/1 hour):
 
 ```bash
 ./gpu_burn -tc 3600
+```
+
+For NVIDIA **GTX** GPUs run it like so (this example will run it for 60 minutes/1 hour):
+
+```bash
+./gpu_burn 3600
 ```
 
 ## Test CPU thermals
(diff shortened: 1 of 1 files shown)


content/hardware-failure.md
```diff
@@ -156,7 +156,7 @@ make
 And now we can run it like so (this example will run it for 60 minutes/1 hour):
 
 ```bash
-./gpu_burn -d 3600
+./gpu_burn -tc 3600
```
leviport, line 159: How does this behave on old GPUs without tensor cores?

thomas-zimmerman: This will crash on cards older than RTX 20X0

leviport: Let's keep both commands then, and explain that `-d` must be used for older GPUs.

jklgrasso: I removed the double switch for GTX GPUs and added another prompt for RTX GPUs with the -tc switch.

JDKarner: Screenshot from 2024-03-21 17-19-32
Screenshot from 2024-03-21 17-21-19
Screenshot from 2024-03-21 17-23-01
Screenshot from 2024-03-21 17-25-00


Just to add my 2¢ 
It is probably best to go with a multifaceted approach and do more variations for less time.
-d uses less power than using no argument other than time. While this may stress parts of the GPU more, it doesn't stress the power delivery components as much which could be where the failure is. 

./gpu_burn  1800
./gpu_burn -d 1800
./gpu_burn -tc 1800
./gpu_burn -d -tc 1800

leviport (approved):

thomas-zimmerman (approved):
looks good

PR #1225, opened 2024-04-16
Update deploy.yml

ahoneybun:
This updates the GitHub Actions since the current ones use nodejs12 while GitHub then forces it to use nodejs16 which both are now EOL anyway.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Setup
         uses: actions/setup-node@v4.0.2
         with:
-          node-version: 14
+          node-version: 20
 
       - name: Cache
         id: cache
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -13,18 +13,18 @@ jobs:
...
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4.1.1
         with:
           fetch-depth: 0
 
       - name: Setup
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v4.0.2
         with:
           node-version: 14
```
macifell, line 23: This probably needs to be updated or removed.

burden (approved):

macifell (approved):
:+1:

PR #1227, opened 2024-04-18
Revert "Update deploy.yml"

macifell:
Reverts system76/docs#1225


burden (approved):

thomas-zimmerman (approved):
Let's get search on the support page working again. We really do need to check on getting node updated to a supported version.

PR #1228, opened 2024-04-24
Update install-ubuntu article for 24.04

ahoneybun:
This updates all of the screenshots and some spring cleaning of super old (18.04 and below) screenshots as well.


jacobgkau:
We need to release the packages from `system76-dev/pre-stable` to `system76-dev/stable` before the PPA instructions at the bottom of the article will work for 24.04. I'll try to find out if we can go ahead and do that.

XV-02:
There are aspects of this still in need of update:
- Grub screenshot needs to update to reflect the change to version 2.12/ changes in menu wording
- The User's first interaction is no longer choosing to try or install Ubuntu, but to select language. This is followed by accessibility features, keyboard options, and network connectivity all before the option to try or install is presented. We should rework the flow of the article to reflect that.
- We have an "Important Note" that references a restore process. This article does not detail a restore process. Text should be changed to remove any possibility of miscommunication the article purpose, while still recommending data backups before any install.
- Not strictly in needed, but we _may_ want a note that the install can be completed entirely offline.
- We may wish to add language indicating that the user should choose "interactive install"
- We should decide on whether we recommend Default or Extended installation. The main difference from a cursory glance is probably that the Extended install included expanded spell check (alternative English Dialect and French spellings for a US English install) as well as Thunderbird and the LibreOffice suite. Install footprint increases about 10% (~1GB) between Default and Extended.
- We should remove language regarding `the box labeled Download updates while installing Ubuntu` as this is no longer an option. 
- We should add language regarding the `third party software` and `additional media format` options. We may wish to recommend both, but I'm not sure on that.
- The `Install Ubuntu` table needs updating. I looks like "Erase Ubuntu" may no longer be present, and I don't know if "Upgrade" is still present either. "Something Else" should be changed to "Manual Installation"
- The screenshot for the decrypt prompt should be closer in aspect ration to all the preceding screenshots. This is both to make the screenshot more intelligible by maintaining the context implied in the earlier images, and to avoid looking like a cropped mobile phone capture.

jacobgkau:
> We should add language regarding the `third party software` and `additional media format` options. We may wish to recommend both, but I'm not sure on that.

I typically recommended neither to users who are about to install the `system76-driver` or `system76-driver-nvidia` package. The additional media formats may be fine to include, but e.g. installing the NVIDIA driver is non-useful and just seems like an additional complication when installing our own NVIDIA driver package.

(If the support team's recent experience is that it's more useful to have it installed initially to help get far enough to install ours, though, then I'd defer to their judgement on that. But nouveau has been "good enough" to get booted to install our driver for a while now, from what I've seen.)

jacobgkau:
Oh, I'm installing 22.04 again for something else, and the proprietary drivers + additional media formats used to be a single checkbox. That makes more sense why I recommended neither.

Now that they're separate, it's probably fine to recommend checking the additional media formats but not the "third party software" (proprietary drivers).

XV-02:
Oh, and additionally note:
- We should omit references to DVDs as install media for Ubuntu. The Ubuntu desktop ISO has been larger than single-layer DVDs since 22.04, and 24.04 is now considerably larger.

ahoneybun:
@XV-02 for the following:

"Grub screenshot needs to update to reflect the change to version 2.12/ changes in menu wording"

I don't think the version really matters but the wording could. Getting it to show up in virt-manager seems to be difficult for some reason.

"The screenshot for the decrypt prompt should be closer in aspect ratio to all the preceding screenshots. This is both to make the screenshot more intelligible by maintaining the context implied in the earlier images, and to avoid looking like a cropped mobile phone capture."

The scaling is more of a virt-manager issue perhaps as well. 

I'll see if I can get those but I don't personally view them as blockers. The other points are great and I'm working on addressing them.

content/install-ubuntu.md
```diff
@@ -55,137 +53,135 @@ Ubuntu may need a special option enabled to install and boot from Ubuntu until t
...
-
 And often will prompt for a restart once complete:
 
-![Updates Complete](/images/install-ubuntu/ubuntu-21.04-updates-finished.png)
+![Updates Complete](/images/install-ubuntu/ubuntu-updater-finished.png)
 
 ### Install System76 Driver
 
-Once you've reinstalled Ubuntu, you'll need to download and install the <u>System76 Driver</u>. Open the <u>Terminal</u> (search <u>Terminal</u> from the Ubuntu (<font-awesome-icon :icon="['fab', 'ubuntu']"></font-awesome-icon>) dash or press <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>T</kbd>), then enter the following commands:
+Once you've installed Ubuntu, you'll need to download and install the <u>System76 Driver</u>. Open the <u>Terminal</u> (search <u>Terminal</u> from the Ubuntu (<font-awesome-icon :icon="['fab', 'ubuntu']"></font-awesome-icon>) dash or press <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>T</kbd>), then enter the following commands:
```
XV-02, line 184: screenshot-2024-05-01-19-57-34

I'm not sure why, but the line breaks in this section are borked.

XV-02 (commented):
A few direct changes, and one additionally ask. At the end of the article we should instruct the user to check for and install updates one last time just to make sure all the system76 packages are present and accounted for.

content/install-ubuntu.md
```diff
@@ -55,137 +53,135 @@ Ubuntu may need a special option enabled to install and boot from Ubuntu until t
...
+
+![Type of Installation](/images/install-ubuntu/install-ubuntu-24.04-6.png)
 
-If all of your files are backed up, or if you are installing from scratch, pick the **Install Ubuntu** option to continue with the installation. Please check the box labeled **Download updates while installing Ubuntu** to ensure that your new installation is fully up to date once the installation is complete.
+The **Default selection** will work for most installations while the **Extended selection** includes applications such as the LibreOffice suite and more. Those applications can be installed once the installation is finished if the **Default selection** is used instead.
 
-![Download Updates](/images/install-ubuntu/ubuntu-install-screen2.png)
+![Applications](/images/install-ubuntu/install-ubuntu-24.04-7.png)
+
+We recommend checking the boxes labeled **Install third-party software for graphics and Wi-Fi hardware** and **Download and install support for additional media formats** but they are not required. In the case of the third-party software you will be installing the NVIDIA driver later in this article if you have NVIDIA hardware for that support.
```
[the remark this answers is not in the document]
ahoneybun: @thomas-zimmerman thoughts on this? I just think defaulting to both would be the best experience.

XV-02:
@ahoneybun This looks good. Once support has a firm decision on `third-party` and `media format` install options, it should be alright to go, so let me know.

ahoneybun:
@XV-02 I updated the article about that section, let me know what you think. This is what @thomas-zimmerman wanted.

content/install-ubuntu.md
```diff
@@ -55,137 +53,143 @@ Ubuntu may need a special option enabled to install and boot from Ubuntu until t
...
 
-If all of your files are backed up, or if you are installing from scratch, pick the **Install Ubuntu** option to continue with the installation. Please check the box labeled **Download updates while installing Ubuntu** to ensure that your new installation is fully up to date once the installation is complete.
+![Applications](/images/install-ubuntu/install-ubuntu-24.04-7.png)
 
-![Download Updates](/images/install-ubuntu/ubuntu-install-screen2.png)
+Let's break down what each of the options offer in regards to hardware and software support:
+
+- **Install third-party software for graphics and Wi-Fi hardware**
+
+This installs drivers (some non-free) for support for Wi-Fi (Wireless) and GPU support mainly NVIDIA as the amdgpu driver for AMD GPU's is in the kernel.
```
jacobgkau, line 86: This is a run-on sentence. A comma is needed before "mainly," and would also be needed before your use of "as" to mean "because." I'd recommend using parenthesis so you don't need so many commas right next to each other.

"Wireless" should not be capitalized in the middle of a sentence.

AMDGPU should be all-caps rather than all-lowercase in this context. Let's rephrase "AMD GPU's" so we don't need to worry about whether that apostrophe is correct or not.

I'd also clarify "in the kernel" by adding the word "included," since a user who needs this article may not know what that means intuitively.

Finally, rephrase "for support for Wi-Fi" so you're not repeating "for" so close together.

> This installs drivers (some non-free) to provide support for Wi-Fi (wireless) and GPU support (mainly NVIDIA, as the AMDGPU driver for AMD graphics is included in the kernel).

content/install-ubuntu.md
```diff
@@ -55,137 +53,143 @@ Ubuntu may need a special option enabled to install and boot from Ubuntu until t
...
+
+- **Download and install support for additional media formats**
+
+This installs packages for playing media files such as MP3, MP4 and other formats that may need non-free codecs for the OS to play the content.
```
jacobgkau, line 90: Please add an oxford comma after MP4.

> This installs packages for playing media files such as MP3, MP4, and other formats that may need non-free codecs for the OS to play the content.

jacobgkau (changes requested):

content/install-ubuntu.md
jacobgkau: Actually, you shouldn't be repeating the word "support: so close together, either.

> This installs drivers (some non-free) to provide Wi-Fi (wireless) and GPU support (mainly NVIDIA, as the AMDGPU driver for AMD graphics is included in the kernel).

content/install-ubuntu.md
```diff
@@ -55,137 +53,143 @@ Ubuntu may need a special option enabled to install and boot from Ubuntu until t
 
 From the menu shown the 'nomodeset' option can be selected during this boot. -->
 
-![grub](/images/install-ubuntu/ubuntu-grub.png)
+![grub](/images/install-ubuntu/GRUB-2.12.png)
 
-Once Ubuntu starts, you will be asked to **Try Ubuntu without installing** or to **Install Ubuntu**. If you are attempting to restore a broken operating system, please choose **Try Ubuntu**.
+Once Ubuntu starts and you have selected your language, any accessibility features that you may need, your keyboard selection, and network connectivity you will be asked to **Try Ubuntu without installing** or to **Install Ubuntu**. If you are attempting to restore a broken operating system, please choose **Try Ubuntu**.
```
jacobgkau, line 58: Can we split the first sentence here so it's not so long? (Otherwise, you would need a comma after "network connectivity.")

> Once Ubuntu starts, you will be asked to select your language, any accessibility features that you may need, your keyboard layout, and network connectivity. Next, you will be asked to **Try Ubuntu without installing** or to **Install Ubuntu**.

content/install-ubuntu.md
jacobgkau: If you're going to call out AMD here, why aren't you also calling out Intel?

Maybe simplify even further to:

> This installs drivers (some non-free) to provide Wi-Fi/Bluetooth and GPU support (mainly for NVIDIA graphics, as AMD and Intel graphics are supported natively by the Linux kernel).

Everyone knows that Wi-Fi means wireless, and nobody reading this article is likely to care what the AMDGPU module is called while performing this installation process.

content/install-ubuntu.md
```diff
@@ -55,7 +55,7 @@ From the menu shown the 'nomodeset' option can be selected during this boot. -->
 
 ![grub](/images/install-ubuntu/GRUB-2.12.png)
 
-Once Ubuntu starts and you have selected your language, any accessibility features that you may need, your keyboard selection, and network connectivity you will be asked to **Try Ubuntu without installing** or to **Install Ubuntu**. If you are attempting to restore a broken operating system, please choose **Try Ubuntu**.
+Once Ubuntu starts and you have selected your language, any accessibility features that you may need, your keyboard selection, and network connectivity. Then you will be asked to **Try Ubuntu without installing** or to **Install Ubuntu**. If you are attempting to restore a broken operating system, please choose **Try Ubuntu**.
```
jacobgkau, line 58: I didn't say to change it to `Then you will be asked`, I said `Next, you will be asked`. A comma is needed after an introductory word like that, the way you have it here is incorrect.

jacobgkau (changes requested):

content/install-ubuntu.md
```diff
@@ -83,11 +83,11 @@ Let's break down what each of the options offer in regards to hardware and softw
 
 - **Install third-party software for graphics and Wi-Fi hardware**
 
-This installs drivers (some non-free) for support for Wi-Fi (Wireless) and GPU support mainly NVIDIA as the amdgpu driver for AMD GPU's is in the kernel.
+This installs drivers (some non-free) for Wi-Fi and GPU support mainly, NVIDIA as the amdgpu driver for AMD GPU's is in the kernel.
```
jacobgkau, line 86: Like I said, the comma would need to go _before_ "mainly," not after. Please update to the suggested wording:

> This installs drivers (some non-free) to provide Wi-Fi/Bluetooth and GPU support (mainly for NVIDIA graphics, as AMD and Intel graphics are supported natively by the Linux kernel).

content/install-ubuntu.md
jacobgkau, line 58: Why did you go with `Then you will be asked` instead of the suggested `Next, you will be asked`?

jacobgkau (changes requested):

jacobgkau:
I'm just going to make a commit myself to solve the rest of the typos.

jacobgkau (approved):
Talked to @XV-02, and we're good to approve this from our end. I gave the article another through read-through and made some additional updates.

thomas-zimmerman (approved):
I really like the updates -- let's get this live.

PR #1229, opened 2024-04-25
Update upgrade-ubuntu article for 24.04

ahoneybun:
This updates all of the screenshots.


XV-02:
This will also need some edits to content:
- Omit references to DVDs as install media. The Ubuntu desktop ISO has exceeded single-layer DVD size since 22.04, and is now considerably larger than a single-layer DVD. 
- In other articles we reference the lower left-hand 'show apps' button. This is present on Ubuntu 22.04 and 24.04. Instead of `Open the Activities Overview by clicking on the Activities button in the top left and then search for Software Updater.` we should reference the 'show apps' button. This is due to Gnome making the activities button unintuitive beginning with version 45, where it transitioned from an immediately identifiable word to ~~an immediately blobby blob of blobiness that communicates blobularity with blobbing aplomb~~ a less identifiable workspace indicator. In contrast, we can establish the 'show apps' paradigm across multiple articles, and benefit from it looking like it might actually be a normal interactable.
- We should highlight that the Updater may run for a moment before `Configure Version Updates` step.

brandonbaez7 (dismissed):

brandonbaez7 (approved):

thomas-zimmerman (approved):
decent update as a first step. With the first point release out, I would like to make sure we have updated docs as most of our Ubuntu 22.04 customers upgrade.

jacobgkau (approved):
I corrected a couple of typos and tried to make some of the grammar more consistent. I'm good with getting this out.

thomas-zimmerman (approved):
Nice polish on the language

PR #1230, opened 2024-04-25
Update ubuntu-basics article for 24.04

ahoneybun:
This updates all of the screenshots and some spring cleaning of super old (18.04 and below) screenshots as well.


XV-02:
Specific notes:
- the second paragraph under `Show Application` references Ubuntu Software, and should be corrected to App Center. 
- The first paragraph in `Ubuntu App Center` references Ubuntu Software, and should be corrected to App Center. Additionally, the icon is no longer an orange briefcase (honestly, I'm not sure it ever was...) , but an orange shopping/carrier bag.
- For `Top Right Menu` it might be worth getting screenshots from a laptop, so that wi-fi, screen brightness, and battery level are evident as menu features. 
- After `Top Right Menu` we should add a section `System Settings` which describes how to access full system setting - that is the 'gear' icon from the Top-Right-Menu and 'settings' application as selected from the activities view or via search.

A couple of general notes:
- We don't ship any products with either an ubuntu or pop branded super-key. We should probably specify `Super+` for shortcuts which leverage the super key, while linking the Ubuntu Keyboard Shortcuts article (which may also need updating)
- We use the terms "launcher" and "dock" in the article, but neither is defined, and they seem to be used interchangeably. In 24.04, this is called the "dash". We should probably call out the dash specifically - with a dedicated screenshot highlighting the dash (dim everything except the dock while also box-highlighting the dash itself) - and then used the term "dash" consistently from there.
- We should highlight - in whichever section it best fits - that applications can be pinned to the dash for quick access.

ahoneybun:
@XV-02 :

"We don't ship any products with either an ubuntu or pop branded super-key. We should probably specify Super+ for shortcuts which leverage the super key, while linking the Ubuntu Keyboard Shortcuts article (which may also need updating)"

We don't **currently** but we have in the past so I'd rather keep it and add the note about the Super key.

It looks like in the Ubuntu docs the Dock is called the Dash?:

https://help.ubuntu.com/lts/ubuntu-help/shell-introduction.html.en

ahoneybun:
@XV-02 we do have "We should highlight - in whichever section it best fits - that applications can be pinned to the dash for quick access." in the launcher section.

jacobgkau (approved):

jacobgkau:
The article referred to the left sidebar as the "launcher" in the first section and the "dock" in all the other sections (it looks like it was already doing that, so it wasn't an issue caused by this PR). I went ahead and changed "launcher" to "dock" in the first section so they all match.

thomas-zimmerman (approved):
LGTM

PR #1232, opened 2024-05-03
Add gnome-shell-extensions to the packages needed for Ubuntu

ryanbaekr:
Ubuntu 24.04 doesn't ship with gnome-shell-extensions


jacobgkau (commented):
Thank you for the update. Are you sure `gnome-shell-extensions` is needed and not just `gnome-shell-extension-prefs`? It looks like the former includes a number of unrelated extensions: https://packages.ubuntu.com/noble/all/gnome-shell-extensions/filelist

ryanbaekr:
> Thank you for the update. Are you sure `gnome-shell-extensions` is needed and not just `gnome-shell-extension-prefs`? It looks like the former includes a number of unrelated extensions: https://packages.ubuntu.com/noble/all/gnome-shell-extensions/filelist

Good catch, I think you are right

jacobgkau (approved):

PR #1234, opened 2024-05-07
Fix typo in launch-keyboard.md

taivlam:
Use "its" (possessive), instead of "it's" (contraction)


ahoneybun (approved):
Thanks for this fix!

leviport (approved):

PR #1236, opened 2024-05-29
Update pop-keyboard-shortcuts.md

trenner1:
add the key strokes for moving active windows between upper and lower displays

### DIFF
--- a/content/pop-keyboard-shortcuts.md
+++ b/content/pop-keyboard-shortcuts.md
@@ -70,12 +70,13 @@ Directional actions can use either the standard arrow keys or their Vim equivale
 | ------------------------------------------------------------------------------------------ | ---------------------------------------------- |
 | <kbd>SUPER</kbd> + <kbd>Ctrl</kbd> + <kbd>↑</kbd>/<kbd>↓</kbd>                               | Navigate between workspaces                    |
 | <kbd>SUPER</kbd> + <kbd>Home</kbd>/<kbd>End</kbd>                                            | Navigate to first/last workspace               |
-| <kbd>SUPER</kbd> + <kbd>Shift</kbd> + <kbd>Direction keys</kbd>                              | Move active window between workspaces/displays |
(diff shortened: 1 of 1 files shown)


content/pop-keyboard-shortcuts.md
```diff
@@ -70,12 +70,13 @@ Directional actions can use either the standard arrow keys or their Vim equivale
...
 | <kbd>SUPER</kbd> + <kbd>Home</kbd>/<kbd>End</kbd>                                            | Navigate to first/last workspace               |
-| <kbd>SUPER</kbd> + <kbd>Shift</kbd> + <kbd>Direction keys</kbd>                              | Move active window between workspaces/displays |
+| <kbd>SUPER</kbd> + <kbd>Shift</kbd> + <kbd>Direction keys</kbd>                              | Move active window between  upper and lower workspaces as well as right and left monitors respectively          |
+| <kbd>SUPER</kbd> + <kbd>Shift</kbd> + <kbd>Ctrl</kbd> + <kbd>↑</kbd>/<kbd>↓</kbd>            | Move the active window between upper and lower displays
 | <kbd>SUPER</kbd> + <kbd>ESC</kbd>                                                            | Lock the screen                                |
 
 ### Use the launcher
 
-The launcher allows searching through open windows and installed applications, and also has the additional functions listed below.
+The launcher allows searching through open windows and installed applications and has the additional functions listed below.
```
jacobgkau, line 79: While I appreciate the drive for conciseness, this change makes the "and"s run together and makes the sentence less clear. I think the old wording for this line was better.

If you want to make it more grammatically correct than it was originally, just add "it" after the "and" (following the comma) since this is an independent clause and should have a subject. (So it should read "and it also has..." after the comma.) Or simply separate it out into two sentences to get rid of the second "and" here.

jacobgkau (changes requested):

content/pop-keyboard-shortcuts.md
jacobgkau: I'll fix it myself since it's been a while and the PR author doesn't seem to be around at the moment.

jacobgkau:
I don't seem to have permission to push to @trenner1's branch. I've opened a new PR with the necessary updates here: https://github.com/system76/docs/pull/1245

PR #1238, opened 2024-06-13
Update cuda.md, replace cuda 12.2 by 12.5

UrielCh:
Cuda 12.2 is deprecated, and display nasty warning.

the driver version 555 is currently 6 day old.

the driver version 550 was built on top of cuda 12.4.


ahoneybun:
Since 555 is not stable yet I believe we would just swap 12.2 with 12.4, is that not correct?

UrielCh:
550 is installed by default out of the box and works with cuda 12.4.

the current PR will be ... fully valid in a few days or week, let me switch to 12.4

UrielCh:
I had replace nvidia-driver-555 by nvidia-driver-550
and cuda 12.5 by 12.4.

that should be a stable version.

thomas-zimmerman (approved):
Thank you for the PR @UrielCh 

Changes look correct for the update to 12.5

jacobgkau (approved):
Tested this on an NVIDIA system & the updated steps work fine. The deprecation notice no longer appears with the new container image.

PR #1239, opened 2024-06-17
add note for EC update

thomas-zimmerman:
Firmware updates that include EC firmware updates require a full power cycle to complete. Added a note suggesting either AC power unplug/plug or holding power button for 15 seconds.


XV-02 (approved):
This tweak provides a layer of discoverability around power resets for EC, and does so in a straight forward manner.

ahoneybun (approved):

PR #1240, opened 2024-06-30
Add note about need to disable ppd on NixOS

spaceoden:
power-profiles-daemon (enabled by default on Gnome NixOS installs) must be disabled for system76-power to start. This PR includes that info in docs.


ahoneybun (approved):
Thank you for the PR! This is similar to a workaround for Arch I believe.

leviport (approved):

PR #1241, opened 2024-07-06
Update the steps to generate a log file in freezing-behavior.md

ws909:
The existing explanation was for a previous version of the system.

### DIFF
--- a/content/freezing-behavior.md
+++ b/content/freezing-behavior.md
@@ -91,7 +91,7 @@ sudo nvme smart-log /dev/nvme0n1
 
 ## Collect Logs
 
-If you are using Pop!\_OS, the Settings application has the ability to generate a robust log file. To generate a log file, open the Settings application, select the "Support" section, and click on the button named "Create Log Files". After you type your password into the prompt that appears, a log file will be generated in your home folder. 
+If you are using Pop!\_OS, the Settings application has the ability to generate a robust log file. To generate a log file, open the Settings application, select the "Support" section, and click on the button named "Create Log Files". After you type your password into the prompt that appears, a log file will be generated in your home folder.
 
 If you are on a System76 computer _not_ running Pop!\_OS, use the System76 Driver to generate the log file. If the driver is not already installed, follow the steps in [this article to install it](https://support.system76.com/articles/system76-driver) before proceeding. To run the driver, access your Activities menu in the top left of your Desktop, type System76 and select the System76 driver. After opening the application, click the button in the bottom right to Create Log Files, then a log file will populate in your home folder.
 
--- a/content/freezing-behavior.md
+++ b/content/freezing-behavior.md
@@ -91,4 +91,4 @@ sudo nvme smart-log /dev/nvme0n1
 
 ## Collect Logs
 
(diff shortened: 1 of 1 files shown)


ahoneybun (changes requested):
I think we should have both and just update the steps for the driver since it also applies to system76 systems.

ahoneybun:
> Do you mean that the old explanation should be kept there? It's already listed on the page linked to at the end. It was moved to the Settings application a few OS versions ago now.
> 
> I have not checked if other help pages also have this outdated explanation; if changes are made, they should probably be made for every help page.

No it is in both the Settings (only on Pop!_OS) and in the System76 Driver for both Ubuntu and Pop!_OS.

ws909:
Is it better to just remove the detailed explanation from the page, then, and rely on the link? For instance, 

>  Please attach a log file in a reply to this ticket. For a detailed list of steps to generate a log file, or if you are having any issues, you can refer to this [article](/articles/log-files).

In all fairness, I do find it a bit strange that this section exists in the first place. I haven't made any contributions here before, so my opinion certainly doesn't carry a lot of weight. I did however find the section quite out of place when I read it. I was looking through a few troubleshooting pages, and it was the first time I heard anything of support tickets. I was not aware of this option, and when looking at the site again, it did not seem to be readily available anywhere (eventually found a link buried in [here](https://support.system76.com/articles/before-you-open-a-support-ticket/)). It would be a reasonable section to include if it was sent by a support agent as a response to an already opened ticket, but it was out of place when looking at that page outside of that context. Other similar troubleshooting pages do not have this section.

ahoneybun:
> Is it better to just remove the detailed explanation from the page, then, and rely on the link? For instance,
> 
> > Please attach a log file in a reply to this ticket. For a detailed list of steps to generate a log file, or if you are having any issues, you can refer to this [article](/articles/log-files).
> 
> In all fairness, I do find it a bit strange that this section exists in the first place. I haven't made any contributions here before, so my opinion certainly doesn't carry a lot of weight. I did however find the section quite out of place when I read it. I was looking through a few troubleshooting pages, and it was the first time[1](#user-content-fn-1-e16d9e7dfd9b32ddfd7136a699fb71fe) I heard anything of support tickets. I was not aware of this option, and when looking at the site again, it did not seem to be readily available anywhere (eventually found a link buried in [here](https://support.system76.com/articles/before-you-open-a-support-ticket/)). It would be a reasonable section to include if it was sent by a support agent as a response to an already opened ticket, but it was out of place when looking at that page outside of that context. Other [similar troubleshooting page](https://support.system76.com/articles/login-loop-pop/)s do not have this section.
> ## Footnotes
> 
>     1. Actually, I see now that the top of the page has "If you have a support ticket open"; I was still a bit confused at the at the section. I may have missed this note while looking through the page and scrolling down. [↩](#user-content-fnref-1-e16d9e7dfd9b32ddfd7136a699fb71fe)

That is for customers who have system76 hardware since they would send the logs to one of the support folks such as myself.

ahoneybun (approved):
I believe this should be good now.

leviport (changes requested):
@ahoneybun can you fix that lint issue? Other than that, this looks good.

leviport (approved):
Thanks!

ahoneybun (approved):

PR #1242, opened 2024-07-24
Add info on docker usage to rocm install instructions

jessecambon:
I ran into some permissions issues using Docker Desktop with ROCM and was able to resolve these issues by using docker.io instead and running docker as sudo. You can refer to the comments [here](https://gist.github.com/guilt/6c901f7ac0a726685b6334798da77c00?permalink_comment_id=5131866#gistcomment-5131866) for more information.

I thought it would be good to add a note here so that others can avoid these issues.

### DIFF
--- a/content/rocm.md
+++ b/content/rocm.md
@@ -88,7 +88,7 @@ system76@pop-os:~$
 
 Instructions from AMD for building and running a docker image with Pytorch and ROCm are [here](https://rocm.docs.amd.com/projects/install-on-linux/en/latest/how-to/3rd-party/pytorch-install.html#using-docker-with-pytorch-pre-installed). Note that you will likely have to run the docker commands as sudo (ie. `sudo docker run -it ...`) for typical docker setups.
 
-Also note that Docker Desktop runs in a virtual environment so it's recommended to use the Docker engine instead to avoid permissions issues. One way to install the Docker engine is to install the `docker.io` package:
+Also note that Docker Desktop runs in a virtual environment so it's recommended to use the Docker engine instead to avoid permissions issues. Instructions on installing the Docker engine are [on Docker's website](https://docs.docker.com/engine/install/ubuntu/). Another way to install the Docker engine is to install the `docker.io` package which is maintained by Ubuntu: 
 
 ```sh
 sudo apt install docker.io
--- a/content/rocm.md
+++ b/content/rocm.md
@@ -84,6 +84,19 @@ Agent 2
 system76@pop-os:~$ 
 ```
 
+### Docker Compatibility
+
+Instructions from AMD for building and running a docker image with Pytorch and ROCM are [here](https://rocm.docs.amd.com/projects/install-on-linux/en/latest/how-to/3rd-party/pytorch-install.html#using-docker-with-pytorch-pre-installed). Note that Docker Desktop runs in a virtual environment so it's recommended to use the docker.io package instead to avoid permissions issues. To install docker.io:
+
+```sh
+sudo apt install docker.io
+sudo usermod -aG docker $USER
+```
+
+Also note that you may need to run docker commands using sudo (ie. `sudo docker run -it ....`). 
+
 ### Blender Compatibility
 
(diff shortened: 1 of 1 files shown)


jacobgkau:
The terminology for "docker.io" (as opposed to Docker Desktop) is Docker Engine or Docker CE: https://docs.docker.com/engine/install/ 

I think it'd make sense to provide a link to Docker's installation instructions as an alternative to the Ubuntu-packaged version, since Docker recommends using their own repo: https://docs.docker.com/engine/install/ubuntu/ 

Aside from that, this looks reasonable. It's expected that Docker commands require sudo; other options such as Podman would allow rootless containers.

jessecambon:
Makes sense. I haven't tried Docker CE, but I assumed docker io was recommended as it's in the CUDA instructions: https://support.system76.com/articles/cuda

jacobgkau:
> I haven't tried Docker CE,

You are using the Ubuntu package for Docker CE.

jessecambon:
@jacobgkau I edited the PR per your comments

jacobgkau (approved):
Thanks for the contribution! The linked AMD tutorial works on pang12.

ahoneybun (approved):

jessecambon:
@jacobgkau [the current ROCm install instructions](https://support.system76.com/articles/rocm/) are for ROCm version 6.1. I'm curious if you have been able to install version 6.2? In theory we should just be able to edit the file`/etc/apt/sources.list.d/rocm.list` and point the apt entry to [one of the version 6.2 URLs](https://repo.radeon.com/rocm/apt/). However, when I do this I run into issues with broken packages (shown below). I'm not sure if this is a ROCm issue or an incompatibility with something in Pop! OS.  

Let me know if there's a better place to ask for help on this. 

```sh
(base) cambonator@pop-os:~$ sudo apt install rocm --upgrade
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 rocm : Depends: rocm-utils (= 6.2.0.60200-66~22.04) but 6.2.1.60201-112~22.04 is to be installed
        Depends: rocm-developer-tools (= 6.2.0.60200-66~22.04) but 6.2.1.60201-112~22.04 is to be installed
        Depends: rocm-openmp-sdk (= 6.2.0.60200-66~22.04) but 6.2.1.60201-112~22.04 is to be installed
        Depends: rocm-opencl-sdk (= 6.2.0.60200-66~22.04) but 6.2.1.60201-112~22.04 is to be installed
        Depends: rocm-ml-sdk (= 6.2.0.60200-66~22.04) but 6.2.1.60201-112~22.04 is to be installed
        Depends: mivisionx (= 3.0.0.60200-66~22.04) but 2.5.0.60100-82~22.04 is to be installed
        Depends: migraphx (= 2.10.0.60200-66~22.04) but 2.10.0.60201-112~22.04 is to be installed
        Depends: rpp (= 1.8.0.60200-66~22.04) but 1.8.0.60201-112~22.04 is to be installed
        Depends: rocm-core (= 6.2.0.60200-66~22.04) but 6.2.1.60201-112~22.04 is to be installed
        Depends: migraphx-dev (= 2.10.0.60200-66~22.04) but 2.10.0.60201-112~22.04 is to be installed
        Depends: mivisionx-dev (= 3.0.0.60200-66~22.04) but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
```

jacobgkau:
@jessecambon Why did you specify `--upgrade` in your `apt install` command?

I'm getting a different error from you when trying to install version 6.2.0 with `6.2` in the URL:

```
The following packages have unmet dependencies:
 rocdecode: Depends: mesa-amdgpu-va-drivers but it is not installable
```

I get the same message with `6.2.2` or `latest` in the URL. Did you deviate from the System76 instructions and manually install a newer `amdgpu-install` .deb package from the AMD instructions?

jessecambon:
@jacobgkau  `--upgrade` probably isn't necessary there. I didn't install amdgpu-install, but I did get the same `mesa-amdgpu-va-drivers` error message you are getting when running `sudo apt install rocm` with `6.2.2` and `latest`. The error message I showed above was from using 6.2.

PR #1243, opened 2024-08-01
Update system-firmware.md

esplinr:
Make clear what happens if the firmware updater is run with external devices plugged in.

### DIFF
--- a/content/system-firmware.md
+++ b/content/system-firmware.md
@@ -21,7 +21,7 @@ These instructions are for System76 owners who have been prompted for a firmware
 ### Before You Begin
 
 * Make sure the laptop is **plugged into the AC Adapter** for the entire firmware updating process. USB Type-C charging is **not** supported during firmware updates on the following models: Galago Pro (galp3-c, galp4), Darter Pro (darp5, darp6)
-* **Disconnect external devices**, including any displays or USB devices. The firmware update will restart and the update will not be written if external devices are connected (no error is shown).
+* **Disconnect external devices**, including any displays or USB devices. If external devices are connected, the firmware update may restart without being written (and without showing an error).
 * **Take a picture of these instructions** or pull up [s76.co/sfu](https://s76.co/sfu) on a phone or other device for reference.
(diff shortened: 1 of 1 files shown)


content/system-firmware.md
```diff
@@ -21,7 +21,7 @@ These instructions are for System76 owners who have been prompted for a firmware
 ### Before You Begin
 
 * Make sure the laptop is **plugged into the AC Adapter** for the entire firmware updating process. USB Type-C charging is **not** supported during firmware updates on the following models: Galago Pro (galp3-c, galp4), Darter Pro (darp5, darp6)
-* **Disconnect external devices**, including any displays or USB devices.
+* **Disconnect external devices**, including any displays or USB devices. The firmware update will restart and the update will not be written if external devices are connected (no error is shown).
```
jacobgkau, line 24: I don't think every single display or USB device will necessarily have this effect. Might need testing, but the wording probably needs to at least say "may" instead of "will."

esplinr: Thanks for the comment. I updated to text.

esplinr:
Is there anything else I need to do for this contribution to be accepted?

jacobgkau (approved):
I don't really think this is necessary (do we need to explain what happens at every step if the user doesn't follow the instructions?), but I guess it could help someone recognize what's wrong if they accidentally leave something plugged in and are only thinking about the symptom.

To try and keep these bullet points from getting so unwieldy people skim them over without really reading them, I've separated the tangential information into sub-bullet points.

jacobgkau:
@ahoneybun You replaced one linter issue with another. Let me take a look at what it needs.

jacobgkau (approved):

ahoneybun (approved):

esplinr:
Thank you for merging. This problem has repeatedly caught me off-guard since I bought my laptop from System76. Hopefully having the hint be explicit will help me remember next time. I agree that it would be much better for the updater to show an error instead of silently restarting.  The current behavior is baffling if you don't remember what a successful update should look like.

PR #1256, opened 2024-10-25
Add article: VirtualBox Black Screen on Threadripper

jacobgkau:
Nested paging must be disabled on some Threadripper and Threadripper Pro products in order for VirtualBox guests to display properly.

I compressed the larger screenshots with WebP, but left the smaller ones as optimized PNG because the compression looks bad when they're scaled up.


leviport (approved):

ahoneybun (approved):
LGTM live

PR #1260, opened 2024-12-07
Update windows.md

ahoneybun:
add addw4 to the table


leviport (approved):

ahoneybun:
@leviport Stetson asked if the Astra has a TPM 2.0 onboard and if so we would want to add it to the table as well no?

leviport:
I don't know. It's your PR.

thomas-zimmerman:
I would strongly suggest that we do not want to add the Astra to any list that suggests that windows would install on the system. While it can be done, it's not something that support would want to try and help a customer get done.

thomas-zimmerman (approved):
LGTM

PR #1263, opened 2024-12-18
addresses issue 1262

ahoneybun:
Adds the Pangolin and Serval to the table, also notes about the bonw15-b as well.


content/open-firmware-systems.md
```diff
@@ -42,6 +42,7 @@ The source code for our Open EC can be found [here](https://github.com/system76/
 | Adder WS (addw3 and newer) | Yes | Yes |
 | Bonobo WS (bonw13 and older) | No | No |
 | Bonobo WS (bonw14 and newer) | Yes | Yes |
+| Bonobo WS (bonw15-b) | No | No |
```
leviport, line 45: I think it would be better to break this down.
- bonw14 (yes,yes)
- bonw15 (yes,yes)
- bonw15-b (no,no)

leviport (changes requested):

content/open-firmware-systems.md
```diff
@@ -55,5 +56,8 @@ The source code for our Open EC can be found [here](https://github.com/system76/
 | Lemur Pro (lemp9 and newer) | Yes | Yes |
 | Oryx Pro (oryp5 and older) | No | No |
 | Oryx Pro (oryp6 and newer) | Yes | Yes |
+| Pangolin (pang12 and newer) | No | No |
```
leviport, line 59: Oh also, this could probably be pang10 and newer. Panp9 is probably too old to be relevant, but I'm sure there are pang10/11's still out there kickin' around.

leviport (dismissed):
Technically I guess just "Pangolin" with no model specificity would be accurate, but going older than pang10 probably doesn't make much sense. Looks good!

thomas-zimmerman (dismissed):
LGTM

thomas-zimmerman (approved):
LGTM live as well.

PR #1266, opened 2025-01-21
Updating the page to be more clear and include Thelio Io v2

ahoneybun:
This PR does the following:

- Adds a link to the guide article rather than per model (some pretty old)
- Adds a photo (and text to make it more clear) for the Thelio Io v2 board

I would want to update the v1 photo from the tech docs but I just see this which mean but less clear.

https://tech-docs.system76.com/models/thelio-b4/img/thelio-io-wiring.webp


leviport (approved):
The captions under the photos feel a bit awkward, but I think they still get the point across. There are other oddities with this article that are unrelated to this PR though, so I think this could be merged now, and a future PR could be opened to polish it up a bit more.

thomas-zimmerman:
I like the change and gives us some place to show the different IO boards.

leviport:
@thomas-zimmerman if you leave an approving review, you'll be able to merge it

thomas-zimmerman (approved):
Approving this PR

PR #1268, opened 2025-01-28
Update battery thresholds article

ahoneybun:
This PR does the following:

- Changes the note to a table about the Pangolin models that do not support either option
- Moves the table to the top about thresholds in general rather than the bottom
- Notes that they do not support Charging Thresholds as well as FlexiCharge
- Adds keywords

### DIFF
--- a/content/laptop-battery-thresholds.md
+++ b/content/laptop-battery-thresholds.md
@@ -22,11 +22,14 @@ The default charging profile for System76 laptops with Open Firmware has a start
 
 To determine if your laptop has Open Firmware or proprietary firmware, see [this article](/articles/open-firmware-systems). (If a system has Open Firmware, then it must also have Open EC to work with charging thresholds.) See [Charging Thresholds](#configuring-charging-thresholds-open-firmware) for Open Firmware systems or [FlexiCharger](#configuring-flexicharger-proprietary-firmware) for proprietary firmware systems.
 
-## Do not support Charging Thresholds or FlexiCharge
+## Charging Thresholds or FlexiCharge support
 
-| Model | Codename |
-|:-----:|:---------:|
-| Pangolin | pang12-pang15 |
+| Model | Codename | Charging Thresholds | FleixCharger |
+|:-----:|:---------:|:---------:|:---------:|
+| Pangolin | pang12 | No | No |
+| Pangolin | pang13 | No | No |
+| Pangolin | pang14 | No | No |
+| Pangolin | pang15 | No | No |
 
 ## Configuring Charging Thresholds (Open Firmware)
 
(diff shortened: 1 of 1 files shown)


content/laptop-battery-thresholds.md
```diff
@@ -20,6 +22,12 @@ The default charging profile for System76 laptops with Open Firmware has a start
 
 To determine if your laptop has Open Firmware or proprietary firmware, see [this article](/articles/open-firmware-systems). (If a system has Open Firmware, then it must also have Open EC to work with charging thresholds.) See [Charging Thresholds](#configuring-charging-thresholds-open-firmware) for Open Firmware systems or [FlexiCharger](#configuring-flexicharger-proprietary-firmware) for proprietary firmware systems.
 
+## Do not support Charging Thresholds or FlexiCharge
+
+| Model | Codename |
+|:-----:|:---------:|
+| Pangolin | pang12-pang15 |
```
leviport, line 29: If you're going to make a table, shouldn't there be a line for each model?

ahoneybun:
How about this?

Screenshot_2025-01-28_08-57-40

content/laptop-battery-thresholds.md
```diff
@@ -20,6 +22,15 @@ The default charging profile for System76 laptops with Open Firmware has a start
 
 To determine if your laptop has Open Firmware or proprietary firmware, see [this article](/articles/open-firmware-systems). (If a system has Open Firmware, then it must also have Open EC to work with charging thresholds.) See [Charging Thresholds](#configuring-charging-thresholds-open-firmware) for Open Firmware systems or [FlexiCharger](#configuring-flexicharger-proprietary-firmware) for proprietary firmware systems.
 
+## Charging Thresholds or FlexiCharge support
+
+| Model | Codename | Charging Thresholds | FleixCharger |
```
leviport, line 27: Fleix -> Flexi

content/laptop-battery-thresholds.md
```diff
@@ -20,6 +22,15 @@ The default charging profile for System76 laptops with Open Firmware has a start
...
+|:-----:|:---------:|:---------:|:---------:|
+| Pangolin | pang12 | No | No |
+| Pangolin | pang13 | No | No |
+| Pangolin | pang14 | No | No |
+| Pangolin | pang15 | No | No |
```
leviport, line 32: I feel like if we're going to have a table with both Flexi and Threshold columns, we should make it a more exhaustive model list. Otherwise, this is no better than simply updating the note that was there before.

leviport (changes requested):

leviport (approved):
Works for me

thomas-zimmerman (approved):
LGTM

PR #1272, opened 2025-02-13
Address issue 1201

ahoneybun:
split 22.04 and 20.04 into their own section as the processes are different.


garrettjwilke (approved):

jacobgkau:
Ok, this was actually just obsoleted altogether in https://github.com/system76/docs/pull/1277 a month later.

PR #1275, opened 2025-02-26
Update install-ubuntu.md

thomas-zimmerman:
Add a "apt upgrade" after the PPA is added.


leviport (changes requested):
Since we removed `sudo apt update` after adding the PPA in https://github.com/system76/docs/pull/1274, shouldn't we do the same here to make it consistent? Either that or remove it in https://github.com/system76/docs/pull/1274 because `sudo apt update` won't hurt anything. Either way, I just feel like they should match.

Issue #1201, opened 2023-12-19
Battery percentage article is out of date.


jonathancross:
This article is out of date: https://github.com/system76/docs/edit/master/content/show-battery-percentage.md

The "Show Battery Percentage" option is now part of the Pop_OS settings.
![image](https://github.com/system76/docs/assets/5115470/c451aae3-8a9d-4ad9-9aa1-323d9e29d9ec)

It is no longer available in `gnome-tweak-tool` either.

ahoneybun:
Thanks for the report @jonathancross ! We're updating this in this PR: https://github.com/system76/docs/pull/1202

jbaldwin:
[https://support.system76.com/articles/show-battery-percentage/](https://support.system76.com/articles/show-battery-percentage/)

This page still seems to be out of date at least for pop os 22.04, I ended up finding [https://help.ubuntu.com/stable/ubuntu-help/power-percentage.html.en](https://help.ubuntu.com/stable/ubuntu-help/power-percentage.html.en) which worked.

PR #1277, opened 2025-03-04
move battery article as it is now incorrect for 22.04 and above

ahoneybun:
This addresses issue https://github.com/system76/docs/issues/1201


leviport (approved):

thomas-zimmerman (approved):
Pages does not show up live

PR #1279, opened 2025-03-06
Add tips about getting gpu-burn to compile

raxod502:
* Creating files manually in the root filesystem should be avoided if it is not necessary. In the case of these instructions, setting the `CUDAPATH` variable when building gpu-burn suffices as a replacement; furthermore, the APT alternatives system automatically creates a symbolic link that avoids the need to hardcode a specific version number here.
* Explain how to pass a different GCC compiler path to the Makefile for the gpu-burn project, to avoid compilation errors that occur with the version of GCC that is shipped with Pop!_OS 20.04 by default.

### DIFF
--- a/content/hardware-failure.md
+++ b/content/hardware-failure.md
@@ -123,39 +123,48 @@ Click the 'Run' button to begin the program.
 
 ### GPU Burn (for NVIDIA GPU's only)
 
-We can also test the GPU by using GPU Burn; first, if we're on Ubuntu, we'll need to install git and CUDA with this command:
+We can also test the GPU by using GPU Burn. First, ensure git and CUDA are installed with this command:
 
 ```bash
-sudo apt install git system76-cuda-latest 
+sudo apt install git nvidia-cuda-toolkit
 ```
 
-Next, we can clone the repository with this command:
+Next, clone the repository and enter its directory with these commands:
 
 ```bash
 git clone https://github.com/wilicc/gpu-burn.git
-```
-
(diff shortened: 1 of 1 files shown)


jacobgkau (commented):
This is an excellent contribution. Providing the CUDA path as a Make argument is a simpler solution than creating a symlink in `/usr/`.

However, while testing this on the Pop!_OS 24.04 alpha, I found that the deprecated `system76-cuda-latest` package is not available there. Our [recommended](https://support.system76.com/articles/cuda) replacement, Ubuntu's `nvidia-cuda-toolkit` package, is available for 24.04, as well as 22.04 and 20.04. On all three of those platforms, installing only that package allows gpu-burn to compile successfully without needing to specify the CUDA path. Therefore, it will be simpler to just switch this article to use that package name instead.

Regarding the GCC version, I replicated the issue requiring installing and specifying a non-default version on 22.04 (which includes GCC 11.4 by default). On 20.04 and 24.04, which include GCC 9.3 and 13.3 by default, respectively, gpu-burn compiled by default without needing to switch the GCC version. With that in mind, I'm going to slightly restructure that information to be more pro-active but also to specify which OS version it applies to.

jacobgkau (approved):

thomas-zimmerman (approved):
Looks good in live testing. A somewhat fresh 22.04 Pop!OS install didn't need the GCC (g++-10) version install for make to just work.

raxod502:
Beautiful, thank you so much! Really appreciate y'all's work, and the improvements to my text.

PR #1281, opened 2025-03-18
note on etcher data collection

kspeakman:
This PR is to add a note about data collection in the recommended Windows imaging tool, balena Etcher.

I was dismayed to find the recommended app Etcher had data collection, and it is on by default. I did not think to check until after I had used it. I think this is important to note for future users. Longer term, a different tool should be recommended because of this.

### DIFF
--- a/content/live-disk.md
+++ b/content/live-disk.md
@@ -122,6 +122,8 @@ Once the flash is complete (should look like the screenshot above), it's time to
 
 Etcher is an open source app for Windows, macOS, and Linux that allows you to burn disk images to USB drives. You can download it at [balena.io/etcher/](https://www.balena.io/etcher/).
 
+> Note: Balena Etcher collects usage statistics by default. You can turn off this data collection in Settings.
+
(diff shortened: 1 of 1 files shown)


kspeakman:
In case the maintainers don't use Windows to verify this, here is a screenshot. This is after I turned it off.

image

ahoneybun:
I think it's important to note that it is anonymous data which also includes errors that the software runs into which is helpful for developers.

kspeakman:
> I think it's important to note that it is anonymous data which also includes errors that the software runs into which is helpful for developers.

It looks like they make a serviceable effort to anonymize user data (file paths, etc) before sending. But there's no guarantee. And since it reaches out to the internet regardless of whether the setting is toggled on or off, user's IP is collected. [ref](https://github.com/balena-io/etcher/issues/2766#issuecomment-531437542)

If needed I can just copy/paste the verbiage from their setting, but I can't speak to its accuracy.

thomas-zimmerman (approved):
This is a good change. I would switch away from Etcher if there was a better simple tool that worked on Mac and Windows.

leviport (approved):

PR #1286, opened 2025-04-04
publish web infra page

burden:
fixes the the redirect that was setup for old tickets


jacobgkau (approved):
That's strange behavior (while hidden, the article can be navigated to from the search box, but can't be directly entered via the URL). This is a necessary change.

I updated the title to title case since it will show up in the list of articles now.

thomas-zimmerman (approved):

PR #1289, opened 2025-05-30
Update boot-menu.md

EricMTodd:
Added a catch-all statement instructing users to try all the FX keys if the ones listed on the boot-splash or in this article don't work.

Changed BIOS to BIOS/UEFI.


thomas-zimmerman (approved):
LGTM; not seeing any issue with the changes in local testing.

PR #1292, opened 2025-07-04
Add meer8, previous models, and meer10 to open firmware list

wackbyte:
I found that non-meer9 models were missing from the table. Let me know if this is inaccurate--meer9 seems to be the only open firmware model from what I saw.


leviport (approved):

thomas-zimmerman (approved):
Changes look good live.

PR #1294, opened 2025-07-04
Fix formatting of "Pop!_OS" in article description

wackbyte:
Before:
image
After:
image


leviport (approved):
Weird that this escape character isn't being hidden, but I can confirm that this change does what we'd want it to. Thanks!

thomas-zimmerman (approved):
Still looks good in live review.

PR #1295, opened 2025-07-04
Update Luanti/Minetest article

wackbyte:
- Minetest rebranded as Luanti ([blog post](https://blog.luanti.org/2024/10/13/Introducing-Our-New-Name/)).
- It would be more accurate to refer to Luanti itself as an "engine" rather than a "game"--[their home page](https://www.luanti.org/) also calls it an "engine".
- It can now be [installed from a PPA](https://www.luanti.org/downloads/), simplifying the installation process greatly.
- Content is now available from [ContentDB](https://content.luanti.org/), which is accessible in-game, further simplifying the installation process.
- Multiplayer still works the same, so there was no need to update that section.


thomas-zimmerman (approved):
Live preview of the "new" page looks fine.

leviport (approved):

leviport:
Thanks!

PR #1298, opened 2025-07-18
Create initial-setup-pop.md

Socialhappiness:
Modifying the install pop article to be more focused on the initial setup, removing content that may confuse people who didn't download and install the iso themselves. Preparing support page for upcoming review on mainstream media targeting windows 10 users switching to Linux.

### DIFF
--- a/content/initial-setup-pop.md
+++ b/content/initial-setup-pop.md
@@ -1,0 +1,77 @@
+---
+title: Pop!_OS Initial Setup
+description: >
+    Get started using Pop!\_OS.
+keywords:
+  - Pop!\_OS
+  - initial setup
+  - Reinstall
+  - Installation
+  - Release
+  - setup
+  - new Linux
+
+facebookImage: /_social/article
+twitterImage: /_social/article
+
+hidden: true
+section: switching
+tableOfContents: true
+---
+
+Welcome to Pop!\_OS! Here's what to expect during initial setup! After the first power on of Pop!\_OS, a series of screens will guide through the setup process.
+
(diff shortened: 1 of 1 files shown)


leviport:
This is the same info from https://support.system76.com/articles/install-pop/#installing-pop_os

What is the purpose of duplicating it into a new hidden article?

Socialhappiness:
The idea is to present the content to a new user who is powering on Pop for the first time (but that didn't install it and doesn't plan to distrohop). A user that uses a web browser and word processor and just wants to know what to expect.
What if we change it to Installer walk-thru as the title? Or something along those lines? Powering on Pop!_OS for the first time?
And Add the initial setup to this same one?
I had it hidden because I didn't want it to be published until after feedback.

jacobgkau:
Re: not duplicating content, if part of the current installation article is moved into a different article verbatim, those steps should ideally be removed from the old article, and it should link to the new one. (At that point, the old one would be pretty short, so it might make sense to just fold it into the USB creation article? And if the new article's going to be renamed to be about installation, just make the changes to the one this content was copied from instead of making a new article so the URL can stay the same?)

Although I'm pretty sure there are some minor differences setting up a System76 machine via OEM mode vs. installing from a USB. This new article still covers the partitioning screen (TryInstallView), which isn't shown in OEM mode. So really, if this is about System76 computer setup, it can't just be copied from the USB installation article. (And a third-party PC user who didn't install themselves would not see the installer, since the installer runs off of the USB, and this new article cuts out the part about how to boot from the USB.)

PR #1300, opened 2025-07-18
move clear-cache-and-cookies.md

Socialhappiness:
propose moving to software troubleshooting section


s76-odoo (approved):

thomas-zimmerman:
I think that would be fine. We get less questions about login and more about not seeing orders after a guest purchase. This article doesn't help there.

jacobgkau:
Ok. I'd like to propose tweaking the wording to make this a little more general if it will live in the Software Troubleshooting section, since the current wording assumes you already know we're talking about our own website specifically. I'll push a commit soon.

PR #1301, opened 2025-08-05
chore(ci): update github actions to latest version

btkostner:
This updates the GitHub actions to the latest version, uses cache everywhere (since it was already used a couple of places), and updates the deploy step to use actions instead of deploying via a git branch. This should improve clone performance a bit by not having an orphan with all of the images stored in git.

This will require you to change the github pages settings to deploy via "Github Actions" instead of "Deploy from a branch", and set the cname to `support.system76.com`.

https://github.com/btkostner/docs/settings/pages


.tool-versions
```diff
@@ -0,0 +1 @@
+nodejs 16.16.0
```
btkostner, line 1: Supports things like `mise`, `asdf`, and the `setup-node` action. This is the latest supported version of node that works (all of the dependencies are pretty out of date)

.github/workflows/ci.yml
```diff
@@ -12,14 +11,23 @@ jobs:
...
 
       - name: Setup Node.js
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v4
+        with:
+          cache: npm
+          node-version-file: .tool-versions
+
+      - id: cache
+        name: Cache
```
btkostner, line 23: Might as well use it everywhere since we already are using it below.

.tool-versions
jacobgkau: We appeared to be using Node.js 18 in our local testing setup (a Nix shell) until today, and are now using Node.js 20. I see that the GitHub workflow used to use 14 and will now use 16. I'm not concerned about the version mismatch causing a problem since it already wasn't matching before, but I am confused if I'm comparing the right numbers or not. (Or maybe some dependencies that don't work with >16 versions aren't necessary for local builds but are for live deployment?)

.github/workflows/ci.yml
```diff
@@ -12,14 +11,23 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
```
erinxocon: Checkout and setup node are both version 6 now. It doesn't look like there are any breaking changes for `checkout@v6`. There are some breaking changes with `setup-node@v6`. Could we try and update these latest versions? If we're already testing and updating this, then might as well go with the newest versions out at the time of release.

erinxocon (approved):

PR #1303, opened 2025-08-15
Update flash-firmware-from-usb.md

EricMTodd:
Fixed a small typo on line 65 that read "coping the file(s)" to "copying the file(s)".


leviport (approved):
Good catch

PR #1308, opened 2025-09-20
Add new dependency to DaVinci Resolve installation

jacobgkau:
A new dependency is required to get Resolve running on Pop!_OS 24.04. This package was installed automatically as a dependency of Xorg on 22.04, but needs to be explicitly installed on 24.04 since Xorg isn't included by default (it's not a dependency of Xwayland).


thomas-zimmerman (approved):
LGTM -- live view has no odd formatting on the update.

PR #1310, opened 2025-09-26
Add another dependency for DaVinci Resolve installation

jacobgkau:
`qtwayland5` is also needed on 24.04 with COSMIC. Not required but also available on 22.04, so still don't need to split the list out.


leviport (dismissed):

jacobgkau:
Looks like the PR's been sniped by a linting error in the new COSMIC keyboard shortcuts article that Carl pushed straight to master. I'll fix that.

leviport (approved):

PR #1337, opened 2025-11-15
Update .iso filenames in live-disk.md

evlim:


jacobgkau (approved):
Not strictly necessary since we have a note that the filename changes periodically and may not be verbatim, but no problems with updating it.

jacobgkau:
Since checksumming unintended files shouldn't be dangerous, I wouldn't be opposed to switching to wildcards instead. I do think there's value in teaching people how to do it manually, but there's also value in making it easier to copy/paste to follow.

PR #1346, opened 2025-12-22
creating a basic nvidia article for customers to reference

Socialhappiness:
Would like to add more troubleshooting eventually but couldn't find an article that strictly focused on installing and removing nvidia so we can start there


thomas-zimmerman (approved):
Good base to work from. Not seeing any odd rendering in NPM live view.

Need to expand on why different versions of the NVIDIA driver would be good to use:
 * RTX 50X0 cards require the GSP firmware and open driver to work
 * GTX 10 series cards do not have support in '580 drivers; use the '470 on older cards
 * CUDA version maximum is defined by NVIDIA driver version.

PR #1352, opened 2026-01-30
Firmware USB article: fix reference to Pop/Ubuntu ISO

leviport:
I believe this line was copied from https://support.system76.com/articles/live-disk, but in this context, it is a .img file being written to the USB, not a Pop or Ubuntu ISO.


AlexanderK-S76 (approved):

thomas-zimmerman (approved):
Looks like a good improvement. No odd changes in the live view.

jacobgkau (approved):

PR #1353, opened 2026-01-30
update firmware via live disk changed

AlexanderK-S76:
Updated the section for updating firmware with a live USB since it no longer reflects the modern live disk you'd download from our website due to the transition to COSMIC.

Also added how to get to GUI firmware tool on COSMIC since it's not currently in settings and corrected the expected size of the EFI partition because it is now 1GB on clean install rather than 512MB.


Socialhappiness (approved):
Looks good!

jacobgkau (approved):

PR #1355, opened 2026-02-04
Update the Minecraft installation instructions

remcohaszing:
The old instructions for installing Minecraft were outdated. Minecraft offers a `.deb` file (but no PPA).


jacobgkau:
We can probably drop the version number from the Pop!_OS section's heading. Minecraft is available through Flathub, which is configured by default on all recent versions of Pop!_OS.

remcohaszing:
Done! How do you feel about the contributor node at the end of the document? I don’t mind, but it feels a bit weird after having changed the instructions significantly.

jacobgkau:
I'm not sure what the long-term plans are for the community section of the articles. For now, I can add your GitHub link to the contributor note, assuming you're okay with that.

You added Pop!_Shop screenshots that are a little outdated now; I'm going to swap those out with COSMIC Store screenshots since it'll be quick to do that.

Really, I'd start to question whether this article even needs to exist when the app's in the store (we don't have "how to install ___" articles for the thousands of other apps on Flathub). It's also in the Snap Store, which means it'd show up in Ubuntu's store out-of-the-box as well, and that covers all of the OS's that System76 ships.

jacobgkau (approved):
This PR can be squashed when merging.

thomas-zimmerman (approved):
No odd formatting when looking at the live view.

Is good to get this out but we may want to archive with it being in flathub. If we want to keep this, it may make more sense to swap the focus to how to game on Pop!OS. Highlighting the trade offs in .deb vs flatpak installs if you are modding would be good to cover. 

Let's get this live.

PR #1360, opened 2026-02-25
Update "Change Desktop Environment Documentation" for 24.04 release

deemoun:
Currently the documentation has confusing 22.04 references that are no longer relevant in some cases.

This fixes it.


thomas-zimmerman (approved):
Thank you for the update. Live view looks good. I would like to get this live but we should update the screen shots of the environments as they look now in 24.04.

leviport (approved):

PR #1372, opened 2026-04-13
fix links to nowhere

Socialhappiness:


erinxocon (approved):

jacobgkau (approved):
Confirmed both of these links function after the fixes.

PR #1376, opened 2026-04-22
Add docs note to enable in Extensions

mkenigs:
On Fedora I had to open the Extensions app and enable Pop Shell after installing with dnf

### DIFF
--- a/content/pop-shell.md
+++ b/content/pop-shell.md
@@ -55,6 +55,8 @@ make local-install
 sudo dnf install gnome-shell-extension-pop-shell
 ```
 
+After installing, open the Extensions app and enable the toggle for `Pop Shell`.
+
 #### For Arch
 
 <u>Pop Shell</u> is packaged in the AUR, but for the keyboard shortcuts to work, you will need to build it from source like we did above.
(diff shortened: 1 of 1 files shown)


content/pop-shell.md
```diff
@@ -55,6 +55,8 @@ make local-install
 sudo dnf install gnome-shell-extension-pop-shell
 ```
 
+After installing, open the Extensions app and enable the toggle for `Pop Shell`.
```
mkenigs: I also rebooted, I don't know if that was necessary or not

jacobgkau: With Wayland, I think it is.

jacobgkau (changes requested):
The Extensions app is not installed by default in Fedora Workstation 44. The instructions need to either include installing it, or use the command line instead.

I'll make the necessary change.

jacobgkau (approved):
Tested working (with my modification to the instructions) on Fedora Workstation 44.

thomas-zimmerman (approved):
No odd formatting from looking at this live. Decent change to allow this to just work without more tools.

mkenigs:
Thank you!

PR #1381, opened 2026-05-12
system76-driver: Update Ubuntu instructions for new S76 repo

leviport:


leviport:
There is probably room for improvement on the note about the warning in App Center. I wasn't quite sure where to go with that.

jacobgkau:
Made some tweaks.

- I think it's better to not have two different links (main GitHub repo and Releases page). People are inevitably going to click the first one, then need to search around for the Releases section, when they could just be clicking the second one right to the Releases section.
- Rephrased the App Center warning info to keep it on-topic.
- Tried to retain the old hint that our NVIDIA packages are maintained by us, and not simply a newer version than what Ubuntu might have.
- In the Install Ubuntu article, where we also would've needed to update this information, I instead changed it to link to the dedicated article to reduce the maintenance burden.
- Adjusted the heading levels and added a horizontal rule to try and make clearer where people coming from the Install Ubuntu article can stop reading.

jacobgkau (approved):

leviport:
Good tweaks, thanks for doing that

leviport:
Bypassing merge rules without slowpport review. They can open another PR if they want to fix anything.

PR #1382, opened 2026-05-15
Fast Startup Disable Instructions updated for W10 and W11

### LINKED ISSUE #1373: Update Windows Article - Fastboot
The Windows 11 Settings app has changed from Windows 10, so the method of reaching the Fastboot setting is a little different from what is listed in the article and could use an update.

Image

Alexander-Kendrick:
Changes the first screenshot and the instructions to using the Run window rather than the Windows 10 settings menu because the Windows 10 settings menu isn't the same as what is in Windows 11. This update changes the docs so they will work regardless of if the user is running Windows 10 or Windows 11.

This fixes #1373


jacobgkau (commented):
In a Windows 11 installation, I see `Change what the power buttons do` available directly in the `Hardware and Sound` section. Windows 10 screenshots appear to also have it there. Can we skip the `Choose a power plan` step, or are you seeing it appear differently?

Image

Alexander-Kendrick:
I'd have to verify with a W11 VM, but if we can cut that step out entirely, it's certainly be further improvement.

content/windows.md
```diff
@@ -524,11 +524,11 @@ You can use that info to find the manufacturer, then search for drivers.
 
 ### Disabling Fast Startup
 
-It is also recommended to disable "Fast Startup" within Windows to avoid problems switching between operating systems. To do this, navigate to Settings -> System -> Power & sleep and click `Additional power settings` in the right sidebar:
+It is also recommended to disable "Fast Startup" within Windows to avoid problems switching between operating systems. To do this, we will need to push the (`SUPER` + R) to open a Run window. You can also use Windows Search to look for the "Run" program and launch it from there if you prefer. Once you have Run open, you will need to type in `control` and push enter to launch the Control Panel.
```
Socialhappiness: remove 'we will need'

Socialhappiness (dismissed):
one small suggestion. Approved on support side when Jacob's suggestion is addressed.

content/windows.md
Alexander-Kendrick: Change made :+1:

jacobgkau (approved):

content/windows.md
jacobgkau: I think I'd already made that change as part of 7930cac4b1c309e0eb03eaadcbf138ca01f73735.

PR #1383, opened 2026-05-19
updated paths to tech docs and zip files since they were 404 and remo…

Socialhappiness:
…ved some verbiage


jacobgkau:
I think we can still have direct downloads so people don't need to figure out GitHub (the three-dot menu is fairly nondescript, and I could see it causing confusion for users). Updated to those links, which all seem to work for me. Would be good for anyone else to confirm they're working.

jacobgkau:
Actually, we don't even need to switch these links to point to GitHub. They just needed `/support` added to the front.

jacobgkau (approved):

PR #1384, opened 2026-05-21
Upgrade Ubuntu: Remove old PPA and driver instructions and link to updated article

leviport:
Stetson pointed out that stale info about the old PPA was still lingering in this article, so I removed it and linked to the system76-driver article. This will hopefully prevent the same situation in the future since there will be one less place where information is duplicated.


thomas-zimmerman (approved):
LGTM; no odd formatting changes in live view

jackpot51 (approved):

PR #1385, opened 2026-06-04
Don't recommend Pop version newer than 22.04

leviport:
Since arsandbox stuff isn't packaged for 24.04 yet, we should probably remove "or newer" from the version recommendation for now.


jklgrasso (approved):

leviport:
Bypassing support review. I'm sure they're fine with it.

PR #1393, opened 2026-06-30
Update bluetooth.md

LayersOfAbstraction:
I could not detect any bluetooth devices and was only able to get this working after following the resolution reported in this [redddit forum](https://www.reddit.com/r/pop_os/comments/10od78v/pop_os_bluetooth_not_working/). 

I would like to also state I have faced similar problems on Windows where unplugging cable after shutdown worked. But with PopOS that wasn't the case and only resolved after waiting 60 seconds. Merging this PR would save folks a lot of troubleshooting as my resolution has been mentioned near the top.

### DIFF
--- a/content/bluetooth.md
+++ b/content/bluetooth.md
@@ -60,6 +60,8 @@ If it's stopped, enable it to auto-start and immediately start it using the foll
 ```bash
 sudo systemctl enable --now bluetooth
 ```
+ 
+Even after being enabled successfully some Pop!_OS users have reported the issue only being resolved after shutting down the PC and unplugging the power cable for at least 1 minute.
 
 ## Using Bluetooth Manager (blueman)
 
--- a/content/bluetooth.md
+++ b/content/bluetooth.md
@@ -60,8 +60,13 @@ If it's stopped, enable it to auto-start and immediately start it using the foll
 ```bash
 sudo systemctl enable --now bluetooth
 ```
- 
-Even after being enabled successfully some Pop!_OS users have reported the issue only being resolved after shutting down the PC and unplugging the power cable for at least 1 minute.
+
+If Bluetooth does not enable after that command, try the following steps:
+
+- Shut down the computer.
+- Unplug the computer.
+- Wait 60 seconds.
+- Plug back in and power on the computer.
 
 ## Using Bluetooth Manager (blueman)
 
--- a/content/bluetooth.md
+++ b/content/bluetooth.md
@@ -63,7 +63,7 @@ sudo systemctl enable --now bluetooth
 
 ## Desktop computer power reset
 
-This solution is not applicable to laptops and in some cases can be resolved via these steps. 
+Bluetooth issues on desktop computers can sometimes be fixed by powering off the machine and fully draining power from the capacitors. This method is not applicable to laptop computers since they have a built-in main battery.
 
 - Shut down the computer.
 - Unplug the computer.
--- a/content/bluetooth.md
+++ b/content/bluetooth.md
@@ -61,7 +61,9 @@ If it's stopped, enable it to auto-start and immediately start it using the foll
 sudo systemctl enable --now bluetooth
 ```
 
-If Bluetooth does not enable after that command, try the following steps:
+## Desktop computer power reset
+
+This solution is not applicable to laptops and in some cases can be resolved via these steps. 
 
 - Shut down the computer.
 - Unplug the computer.


LayersOfAbstraction:
FYI I'm on Pop!_OS 24.04 when I enountered the problem. Not sure if you even need further hardware/software info at this point.

content/bluetooth.md
```diff
@@ -60,6 +60,8 @@ If it's stopped, enable it to auto-start and immediately start it using the foll
 ```bash
 sudo systemctl enable --now bluetooth
 ```
+ 
+Even after being enabled successfully some Pop!_OS users have reported the issue only being resolved after shutting down the PC and unplugging the power cable for at least 1 minute.
```
Socialhappiness: Thank you for contributing! What if we worded it like this? 

If Bluetooth does not enable after that command, try the following steps:
- Shut down the computer
- Unplug the computer
- Wait 60 seconds
- Plug back in and power on the computer

Socialhappiness (changes requested):

content/bluetooth.md
LayersOfAbstraction: Ok sure I've added your changes.

Accidentally synced my fork and pulled in an unrelated PR's changes. I've force-pushed to reset the branch to only contain the suggested new bluetooth.md changes. 

See [0ff92b8](https://github.com/system76/docs/commit/0ff92b8e35e38db580c77cd5189c81eabdc48b97). I hope this gets approved soon, thank you.

Socialhappiness (approved):
great! Thank you for contributing!

LayersOfAbstraction:
> great! Thank you for contributing!

Hi Emma,

FYI: [Github deletes workflow runs awaiting approval after 30 days](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/approve-runs-from-forks). I made this PR on June 30th. It's now July 27th. Could you please get my PR peer approved before the 30th? Thank you.

Jordan

leviport:
That isn't anything to worry about.

content/bluetooth.md
```diff
@@ -61,6 +61,13 @@ If it's stopped, enable it to auto-start and immediately start it using the foll
 sudo systemctl enable --now bluetooth
 ```
 
+If Bluetooth does not enable after that command, try the following steps:
```
leviport: A couple things:
- This only applies desktop computers, but this doesn't make that clear.
- This is a pretty drastic step that is not guaranteed to fix the problem, and is very hardware-dependent. I'd rather couch it in language along the lines of "in some cases" or "this might help", rather than inserting it as the next troubleshooting step that everyone should try. I'm not sure how to state that sentiment cleanly though.

leviport (changes requested):

content/bluetooth.md
LayersOfAbstraction: Maybe you you solved much of the problem. 

> This only applies desktop computers, but this doesn't make that clear.

Good catch. A gaming laptop would not get the same resolution. I can mention that. 

> This is a pretty drastic step that is not guaranteed to fix the problem, and is very hardware-dependent.

True it is not guaranteed. How about this sentiment and heading instead so we can ensure only desktop users attempt it?

## Desktop computer power reset

This solution is not applicable to laptops and in some cases can be resolved via these steps. 

- Shut down the computer.
- Unplug the computer.
- Wait 60 seconds.
- Plug back in and power on the computer.

LayersOfAbstraction (commented):
Thanks for your patience. I have added these changes. I hope this looks good. Kindly approve.

content/bluetooth.md
LayersOfAbstraction: FYI @leviport how about this? [89a7fac](https://github.com/system76/docs/pull/1393/commits/89a7faca54e58824258a1eee0b10c66295500ab3)

content/bluetooth.md
```diff
@@ -61,6 +61,15 @@ If it's stopped, enable it to auto-start and immediately start it using the foll
 sudo systemctl enable --now bluetooth
 ```
 
+## Desktop computer power reset
+
+This solution is not applicable to laptops and in some cases can be resolved via these steps.
```
leviport: I think expanding this a bit would help to explain the section better to the reader. Let's change this sentence to:

> Bluetooth issues on desktop computers can sometimes be fixed by powering off the machine and fully draining power from the capacitors. This method is not applicable to laptop computers since they have a built-in main battery.

leviport (changes requested):
Small change, then I think this should be good.

content/bluetooth.md
LayersOfAbstraction: Updated as suggested. Thanks!

leviport (approved):

leviport:
Thanks!

